Issue #2, opened 2019-06-06
bcl2fastq.index_summary: Add documentation


slazicoicr:
The documentation should follow https://www.sphinx-doc.org/en/1.5/ext/example_google.html

Issue #1, opened 2019-06-06
bcl2fastq.index_summary : Disable 'clear value' option for Run drop down menu


slazicoicr:
Having no run selected in the drop down menu makes no sense and causes exception. Remove that option (remove the X in the top right corner).

![image](https://user-images.githubusercontent.com/37381890/59037281-f8800480-883e-11e9-8b29-c0fb2c5c3536.png)

Issue #11, opened 2019-06-07
dashi: Gracefully fail if expected columns are not found


slazicoicr:
All Dashi apps have column names hard coded in them and will crash due to an exception if a column is not found. The column names come from ETL. As ETL is actively developed, it makes no promises about column naming or presence at this point. 

Write a common functionality for apps to gracefully deal with changes in column names or existence. Ideally this should include a clear visual display to the user informing them why the visualization is not working and a message to the developer pinging them about the error.

ETL aims to provide a constants.Column Enum for each module specifying which columns are present. This can be queried before the visualizer starts up.

Given #10, this implementation should also be able to gracefully deal with changes in column names.

PR #13, opened 2019-06-11
Annotations and Clearable bug #1

ijafri26:
#1 
#2

### DIFF
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -10,7 +10,7 @@ index = gsiqcetl.bcl2fastq.parse.load_cache(
     gsiqcetl.bcl2fastq.parse.CACHENAME.SAMPLES,
     './data/bcl2fastq_cache.hd5'
 )
-
+#set variable index(referring to all indices of the run) to data file in module 
 unknown = gsiqcetl.bcl2fastq.parse.load_cache(
     gsiqcetl.bcl2fastq.parse.CACHENAME.UNKNOWN,
     './data/bcl2fastq_cache.hd5'
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -18,19 +18,31 @@ all_runs = index['Run'].sort_values(ascending=False).unique()
 
 all_runs = [{'label': r, 'value': r} for r in all_runs]
 
-# Within the main layout division there are 4 groups:
-#           1. the section with the 100% bar graph
-#           2.  Nested subdivision of:
-#                   a) pie graph at 24%
-#                   b) bar graph at 76%
-#           3. Sample Run hidden
-#           4. Pruned  Unknown Hidden
-# Note: span can be used for inline setting
+""" Sample_run_hidden holds json split format of "Known"
+        columns: "FlowCell","Index1","Index2","LIMS IUS SWID","LaneClusterPF","LaneClusterRaw",
+                 "LaneNumber","LaneYield","QualityScoreSum","ReadNumber","Run","RunNumber","SampleID",
+                 "SampleName","SampleNumberReads","SampleYield","TrimmedBases","Yield","YieldQ30"
+                   
+
+     pruned_unknown_hidden holds json split format of "unknown"
+        columns:  "Count","LaneNumber","Index1","Index2","Run","LIMS IUS SWID
+               
+         """
+
+"""Within the main layout division there are 4 groups:
+          1. the section with the 100% bar graph
+          2.  Nested subdivision of:
+                  a) pie graph at 24%
+                  b) bar graph at 76%
+          3. Sample Run hidden
+          4. Pruned  Unknown Hidden
+            Note: span can be used for inline setting
 
 
-#           Dropdown option's label set to all_runs(combination of unknown and known)
-#           Dropdown multi set to false (default) - only single select possible
-#                 -  multiple select gives error due to inconsistent values
+          Dropdown option's label set to all_runs(combination of unknown and known)
+          Dropdown multi set to false (default) - only single select possible
+                -  multiple select gives error due to inconsistent values
+          """
 layout = html.Div(children=[
     dcc.Dropdown(
         id='run_select',
@@ -63,18 +75,9 @@ layout = html.Div(children=[
         ),
         html.Div(
             [dcc.Graph(id='unknown_index_bar')],
-            style={'width': '76%', 'display': 'inline-block', 'float': 'center'}
+            style={'width': '76%', 'display': 'inline-block', 'float': 'right'}
         ),
     ]),
-    #           Sample_run_hidden holds json split format of "Known"
-    #               columns: "FlowCell","Index1","Index2","LIMS IUS SWID","LaneClusterPF","LaneClusterRaw",
-    #                   "LaneNumber","LaneYield","QualityScoreSum","ReadNumber","Run","RunNumber","SampleID",
-    #                   "SampleName","SampleNumberReads","SampleYield","TrimmedBases","Yield","YieldQ30"
-    #               Index: 6761 - 8307
-
-    #           pruned_unknown_hidden holds json split format of "unknown"
-    #               columns:  "Count","LaneNumber","Index1","Index2","Run","LIMS IUS SWID"
-    #               Index: 218463 - 227804
 
     html.Div(id='sample_run_hidden', style={'display': 'none'}),
     html.Div(id='pruned_unknown_hidden', style={'display': 'none'}),
@@ -96,23 +99,23 @@ except ModuleNotFoundError:
     [dep.Input('run_select', 'value')]
 
 )
-#           Replace run variable iteratively
 def update_sample_run_hidden(run_alias):
     """ When input (run_select) is changed, function is rerun to get json data for selected run
 
         Attributes:
-            run(list): run number
 
         Parameters:
-            run_alias(list): alias of the run
+            run_alias(str): alias of the run
 
         Returns:
              json object with columns, indices, and data, from run variable
+             updates json string "sample_run_hidden"
 """
     run = index[index['Run'] == run_alias]
     run = run[run['ReadNumber'] == 1]
     run = run[~run['SampleID'].isna()]
     run = run.drop_duplicates(['SampleID', 'LaneNumber'])
+
     return run.to_json(date_format='iso', orient='split')
 
 
@@ -125,12 +128,13 @@ def update_pruned_unknown_hidden(run_alias):
         update_pruned_unknown_hidden to get json data for selected run
 
         Attributes:
-            pruned(list): select run_alias'
+            pruned(str): holds function with parameters (run_alias, index, unknown)
         Parameters:
-            run_alias(list): alias of the run
+            run_alias(str): alias of the run
 
         Returns:
              json object with columns, indices, and data, from pruned variable
+             updates json string "pruned_unknown_hidden"
     """
     pruned = gsiqcetl.bcl2fastq.utility.prune_unknown_index_from_run(
         run_alias, index, unknown
@@ -146,13 +150,13 @@ def update_known_index_bar(run_json):
     """ When input (sample_run_hidden) is changed, function is rerun to get update bar graph of sample indices
 
            Attributes:
-               run(list): name of run
-               data(list): empty list appended with values from "For" function
+               data(list): hold values for graphing known (sample) indices
            Parameters:
-               run_json(list): json format of run variable
+               run_json(str): json format of run data
 
            Returns:
-                stacked bar graph of sample indices
+              data and values for the layout of stacked bar graph of sample indices
+              updates bar graph "known_index_bar"
        """
     run = pandas.read_json(run_json, orient='split')
     run['library'] = run['SampleID'].str.extract(
@@ -191,24 +195,24 @@ def update_known_index_bar(run_json):
     [dep.Input('pruned_unknown_hidden', 'children')]
 )
 def update_unknown_index_pie(run_json):
-    """ When input (sample_run_hidden) is changed, function is rerun to get update bar graph unknown indices
+    """ When input (sample_run_hidden) is changed, function is rerun to get updated bar graph of known and unknown indices
 
                Attributes:
-                   pruned(list): select run_alias'
-                   data(list): empty list appended with parameters of "For" function
+                   pruned(str): holds first 30 unknown indices
+                   data(list): holds values to create pie chart to compare unknown and known indices
                Parameters:
-                   run_json(list): json format of run variable
+                   run_json(str): json format of run data
 
-               Returns: stacked bar graph of descending 30 unknown indices
-
+               Returns:
+                   data and layout values for stacked bar graph for unknown indices
+                   updates unknown_index_bar bar graph
            """
     pruned = pandas.read_json(run_json, orient='split')
     pruned['index'] = pruned['Index1'].str.cat(
         pruned['Index2'].fillna(''), sep=' '
     )
     pruned = pruned.sort_values('Count', ascending=False)
-    pruned = pruned.head(30)  # Amount of data visible in graph, largest 30 shown
-
+    pruned = pruned.head(30)
     data = []
     for lane, d in pruned.groupby('LaneNumber'):
         data.append({
@@ -217,7 +221,6 @@ def update_unknown_index_pie(run_json):
             'type': 'bar',
             'name': lane,
         })
-
     return {
         'data': data,
         'layout': {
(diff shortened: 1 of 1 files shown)


bcl2fastq/index_summary.py
```diff
@@ -89,14 +111,17 @@ def update_pruned_unknown_hidden(run_alias):
...
-    run['library'] = run['SampleID'].str.extract(
-        'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_'
+        # convert a Json string into a panda object where orient is the expected format of json 
+    run['library'] = run['SampleID'].str.extract(   # create DataFrame via str.extract 
+        'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_' 
     )
+        #replace index library with index sampleID as a string
     run['index'] = run['Index1'].str.cat(
-        run['Index2'].fillna(''), sep=' '
+        run['Index2'].fillna(''), sep=' ' #concatenate the strings with a space in the middle
```
callunity: Can probably leave this out

bcl2fastq/index_summary.py
```diff
@@ -24,10 +27,16 @@
         id='run_select',
         options=all_runs,
         value=all_runs[0]['value'],
+        clearable=False
     ),
+#           Dropdown option's label set to all_runs(combination of unknown and known)
```
callunity: By convention, comments are often added above the thing that they refer to, to give context before reading the code. Could this comment be moved above the `layout = ...` declaration?

bcl2fastq/index_summary.py
```diff
@@ -89,14 +111,17 @@ def update_pruned_unknown_hidden(run_alias):
...
 def update_known_index_bar(run_json):
     run = pandas.read_json(run_json, orient='split')
-
-    run['library'] = run['SampleID'].str.extract(
-        'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_'
+        # convert a Json string into a panda object where orient is the expected format of json 
+    run['library'] = run['SampleID'].str.extract(   # create DataFrame via str.extract 
+        'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_' 
     )
+        #replace index library with index sampleID as a string
```
callunity: I'm not sure that this is quite accurate, but perhaps Savo can clarify

bcl2fastq/index_summary.py
```diff
@@ -40,31 +49,43 @@
...
     import dash
-    app = dash.Dash(__name__)
+    app = dash.Dash(__name__) 
     app.layout = layout
 
 
 @app.callback(
     dep.Output('sample_run_hidden', 'children'),
     [dep.Input('run_select', 'value')]
+#           Whenever input(drop down) changes, app.callback re-runs and updates
```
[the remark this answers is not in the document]
slazicoicr: Functions should be annotated with docstring. #2 has a link to the format that should be used. A good example is

```
def function_with_types_in_docstring(param1, param2):
    """Example function with types documented in the docstring.

    `PEP 484`_ type annotations are supported. If attribute, parameter, and
    return types are annotated according to `PEP 484`_, they do not need to be
    included in the docstring:

    Args:
        param1 (int): The first parameter.
        param2 (str): The second parameter.

    Returns:
        bool: The return value. True for success, False otherwise.

"""
```

The first sentence should be a succinct explanation of what the function does. The next paragraph adds any necessary details/considerations/limitations.  The `Args` fields are used to describe the parameters that go into the function (for Dash callbacks: what causes this function to be executed) and the `Returns` field explains what the function returns (for Dash Callbacks: what does this function change).
```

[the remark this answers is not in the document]
Comment (callunity) [reply bcl2fastq/index_summary.py]: `The next paragraph adds any necessary details/considerations/limitations.` <- since Irene wasn't the one to write and debug the functions, I don't know that she has the necessary context to add these for this file. The rest should be easier to add though.

File: bcl2fastq/index_summary.py (hunk 3 above)
Comment (slazicoicr) [reply bcl2fastq/index_summary.py]: This takes the `SampleID` column, extracts the library string out of each row value, and assigns those library strings to the new `library` column. 

As an example, from `SWID_11040179_NOVA_0003_Pb_R_PE_346_EX_NoGroup_171017_A00169_0049_AH53FCDMXX` the library string `NOVA_0003_Pb_R_PE_346_EX` gets extracted.

File: bcl2fastq/index_summary.py [hunk 5]
```diff
@@ -89,14 +111,17 @@ def update_pruned_unknown_hidden(run_alias):
...
+    run['library'] = run['SampleID'].str.extract(   # create DataFrame via str.extract 
+        'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_' 
     )
+        #replace index library with index sampleID as a string
     run['index'] = run['Index1'].str.cat(
-        run['Index2'].fillna(''), sep=' '
+        run['Index2'].fillna(''), sep=' ' #concatenate the strings with a space in the middle
     )
-
+    run = run.sort_values('SampleNumberReads', ascending=False)
```
slazicoicr: Why the sort?

The `SampleNumberReads` values are added together at 

https://github.com/oicr-gsi/dashi/blob/294ca750dda09a50824b681cf473e137d79fe32e/bcl2fastq/index_summary.py#L130

and the order should not affect final result.

bcl2fastq/index_summary.py
```diff
@@ -89,14 +110,17 @@ def update_pruned_unknown_hidden(run_alias):
 )
 def update_known_index_bar(run_json):
     run = pandas.read_json(run_json, orient='split')
-
-    run['library'] = run['SampleID'].str.extract(
-        'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_'
+        # convert a Json string into a panda object where orient is the expected format of json
```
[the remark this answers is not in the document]
slazicoicr: A resource on the relationship between Pandas DataFrame and JSON object: https://datatofish.com/export-pandas-dataframe-json/

bcl2fastq/index_summary.py
```diff
@@ -40,31 +49,43 @@
...
     app.layout = layout
 
 
 @app.callback(
     dep.Output('sample_run_hidden', 'children'),
     [dep.Input('run_select', 'value')]
+#           Whenever input(drop down) changes, app.callback re-runs and updates     
 )
-def update_sample_run_hidden(run_alias):
+def update_sample_run_hidden(run_alias):        #Unsure
```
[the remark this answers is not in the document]
ijafri26: Yes, it was a ToDo for myself, have removed it

bcl2fastq/index_summary.py
```diff
@@ -5,100 +5,133 @@
...
 
 @app.callback(
+
     dep.Output('sample_run_hidden', 'children'),
     [dep.Input('run_select', 'value')]
-#           Whenever input(drop down) changes, app.callback re-runs and updates     
+
 )
-def update_sample_run_hidden(run_alias):        
+#           Replace run variable iteratively
```
callunity: Can this be incorporated into the docstring below, so there's only one place where the method is documented?

bcl2fastq/index_summary.py
```diff
@@ -5,100 +5,133 @@
...
+def update_sample_run_hidden(run_alias):
+    """ When input (run_select) is changed, function is rerun to get json data for selected run
+
+        Attributes:
+            run(list): run number
```
callunity: Since `run` gets reassigned many times in this function each time serving a different purpose, I think you could exclude documenting it.

bcl2fastq/index_summary.py
```diff
@@ -5,100 +5,133 @@
...
+
+        Parameters:
+            run_alias(list): alias of the run
```
callunity: I believe `run_alias` is of type `str` here, can you confirm? (Same for methods below)

bcl2fastq/index_summary.py
```diff
@@ -5,100 +5,133 @@
...
 @app.callback(
     dep.Output('pruned_unknown_hidden', 'children'),
     [dep.Input('run_select', 'value')]
 )
 def update_pruned_unknown_hidden(run_alias):
+    """ When input (run_select) is changed, function prune_unknown_index_from_run is run within
+        update_pruned_unknown_hidden to get json data for selected run
+
+        Attributes:
+            pruned(list): select run_alias'
```
callunity: I'm a bit unclear on what `select run alias'` means, could you clarify or remove it? (Same with method below)

bcl2fastq/index_summary.py
```diff
@@ -110,18 +143,24 @@ def update_pruned_unknown_hidden(run_alias):
     [dep.Input('sample_run_hidden', 'children')]
 )
 def update_known_index_bar(run_json):
+    """ When input (sample_run_hidden) is changed, function is rerun to get update bar graph of sample indices
+
+           Attributes:
+               run(list): name of run
+               data(list): empty list appended with values from "For" function
```
callunity: Rather than describe the transitions of the variable ("empty list", "appended"), can you describe the final purpose of the variables? Maybe some thing like `object of run data for graphing the indices`. (Same with method below)

bcl2fastq/index_summary.py
```diff
@@ -110,18 +143,24 @@ def update_pruned_unknown_hidden(run_alias):
...
+           Parameters:
+               run_json(list): json format of run variable
```
callunity: probably a `str` format instead of a `list`, and it's probably more accurate to call it `run data`; calling it `run variable` makes it sound like `run_json` depends on `run` existing already, when the opposite is true. (Same as method below)

bcl2fastq/index_summary.py
```diff
@@ -152,12 +191,23 @@ def update_known_index_bar(run_json):
     [dep.Input('pruned_unknown_hidden', 'children')]
 )
 def update_unknown_index_pie(run_json):
+    """ When input (sample_run_hidden) is changed, function is rerun to get update bar graph unknown indices
```
callunity: This docstring says it causes a bar graph to change, but to me it looks like the name of the method implies that a pie chart changes instead. Can these be harmonized?

bcl2fastq/index_summary.py
```diff
@@ -152,12 +191,23 @@ def update_known_index_bar(run_json):
...
+
+               Attributes:
+                   pruned(list): select run_alias'
+                   data(list): empty list appended with parameters of "For" function
+               Parameters:
+                   run_json(list): json format of run variable
+
+               Returns: stacked bar graph of descending 30 unknown indices
```
callunity: It looks to me like this returns data and layout information, but I think it's the a separate method (part of the dashi library) that converts that data into a stacked bar graph.

bcl2fastq/index_summary.py
```diff
@@ -187,6 +237,15 @@ def update_unknown_index_pie(run_json):
      dep.Input('pruned_unknown_hidden', 'children')]
 )
 def update_pie_chart(run_alias, known_json, unknown_json):
+    """ When inputs are changed function os rerun to get updated pie chart
```
callunity: `os` -> `is`

bcl2fastq/index_summary.py
```diff
@@ -187,6 +237,15 @@ def update_unknown_index_pie(run_json):
...
+            Attributes:
+                known(list): json format of known indices
+                pruned(list): json format of unknown indices
```
callunity: For `known` and `pruned`: probably `str` type, and can remove the `json format of`

bcl2fastq/index_summary.py
```diff
@@ -5,66 +5,110 @@
...
             [dcc.Graph(id='unknown_index_bar')],
-            style={'width': '74%', 'display': 'inline-block', 'float': 'right'}
+            style={'width': '76%', 'display': 'inline-block', 'float': 'center'}
         ),
     ]),
+    #           Sample_run_hidden holds json split format of "Known"
+    #               columns: "FlowCell","Index1","Index2","LIMS IUS SWID","LaneClusterPF","LaneClusterRaw",
+    #                   "LaneNumber","LaneYield","QualityScoreSum","ReadNumber","Run","RunNumber","SampleID",
+    #                   "SampleName","SampleNumberReads","SampleYield","TrimmedBases","Yield","YieldQ30"
+    #               Index: 6761 - 8307
```
slazicoicr: Unlike the Columns, the Index range is not fixed. As the input DataFrame grows, the Index upper bound will change.

bcl2fastq/index_summary.py
```diff
@@ -5,66 +5,110 @@
...
+
+    #           pruned_unknown_hidden holds json split format of "unknown"
+    #               columns:  "Count","LaneNumber","Index1","Index2","Run","LIMS IUS SWID"
+    #               Index: 218463 - 227804
+
```
slazicoicr: I would put DataFrame column description in the module level documentation (docstring at the top of the file). The consistency of Column names is critical (see #11), so having that info at the top makes sense to me.

bcl2fastq/index_summary.py
```diff
@@ -5,66 +5,110 @@
...
+    """ When input (run_select) is changed, function is rerun to get json data for selected run
+
+        Attributes:
+            run(list): run number
+
+        Parameters:
+            run_alias(list): alias of the run
+
+        Returns:
+             json object with columns, indices, and data, from run variable
```
slazicoicr: In the `Returns` field, also specify what this function changes. In this case, the hidden div JSON string is updated. This applies to the subsequent annotations as well.

bcl2fastq/index_summary.py
```diff
@@ -174,16 +258,16 @@ def update_pie_chart(run_alias, known_json, unknown_json):
...
+               'data': [{
+                   'labels': ['Known', 'Unknown'],
+                   'values': [known_count, pruned_count],
+                   'type': 'pie',
+                   'marker': {'colors': ['#349600', '#ef963b']},
+               }],
+               'layout': {
+                   'title': 'Flow Cell Composition of Known/Unknown Indices'
+               }
+           }, 'Predicted clusters / produced clusters: {}%'.format(
```
slazicoicr: I think this is over-indented. Did the PEP8 autoformatter do this?

bcl2fastq/index_summary.py
callunity: I would have put that at the top in the description, not at the bottom. From what I can see "updating the hidden div JSON string" is a side effect of this method, and not actually part of the return value, and there's no indication in the code itself (other than the name of the function) that this happens.

slazicoicr: The `@app.callback` decorator just above the function definition makes these connections explicit.

The function parameters are values in the Dash Layout that have just changed and the return value will become the new value in the Dash Layout.

@ijafri26 I will be happy with either convention you pick (in function description or part of Returns description), but be consistent.

bcl2fastq/index_summary.py
ijafri26: yes, I used the pycharm formatter

bcl2fastq/index_summary.py
```diff
@@ -5,66 +5,110 @@
...
                  ),
+
              )],
-            style={'width': '24%', 'display': 'inline-block'}
+            style={'width': '24%', 'display': 'inline-block', }
         ),
         html.Div(
             [dcc.Graph(id='unknown_index_bar')],
-            style={'width': '74%', 'display': 'inline-block', 'float': 'right'}
+            style={'width': '76%', 'display': 'inline-block', 'float': 'center'}
```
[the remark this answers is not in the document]
ijafri26: I was trying to see if it would change the overlapping that occurs if you expand the text box in the bottom left, but forgot to change it back,I can do that in the next commit

bcl2fastq/index_summary.py
ijafri26: I'm not sure why the name has the word "pie", the data actually does return the bar graph at the bottom of the report

bcl2fastq/index_summary.py
callunity: Sure, that makes sense to me. @ijafri26, could you please tweak some of the `Returns` descriptions? Instead of `updates json string "sample_run_hidden"`, which says what the code does, can you describe the effect that this has on the page? You already do this for some of them below, like `updates bar graph "known_index_bar"`.

bcl2fastq/index_summary.py
```diff
@@ -162,9 +240,26 @@ def update_unknown_index_pie(run_json):
...
+                unknown_json: json format of unknown indices data
+            Returns:
+                updated pie chart "known_unknown_pie" with known and unknown indices ratio over total cluster
+                updates value of known_fraction
+    """
     known = pandas.read_json(known_json, orient='split')
     pruned = pandas.read_json(unknown_json, orient='split')
-
+    print(type(known))
+    print(type(pruned))
```
callunity: Stray `print` functions?

bcl2fastq/index_summary.py
```diff
@@ -5,70 +5,117 @@
...
+
     app = dash.Dash(__name__)
     app.layout = layout
 
 
 @app.callback(
+
     dep.Output('sample_run_hidden', 'children'),
     [dep.Input('run_select', 'value')]
+
```
slazicoicr: Spaces were added to this decorator, but not others. My style is not to have spaces between decorator and function.

bcl2fastq/index_summary.py
```diff
@@ -5,70 +5,117 @@
...
 )
 def update_sample_run_hidden(run_alias):
+    """ When input (run_select) is changed, function is rerun to get json data for selected run
+
+        Attributes:
```
slazicoicr: This applies to all function docstrings: remove the Attributes field. If you feel it necessary to document a function variable, do it in the function using the `#` comment.

bcl2fastq/index_summary.py
callunity: Please remove these print functions.

bcl2fastq/index_summary.py
```diff
@@ -127,13 +185,21 @@ def update_known_index_bar(run_json):
     [dep.Input('pruned_unknown_hidden', 'children')]
 )
 def update_unknown_index_pie(run_json):
```
slazicoicr: Almost there. I would update this function name to reflect the fact that it deals with a bar graph rather than a pie chart.

Issue #4, opened 2019-06-06
bcl2fastq.index_summary: Crash when 180810_M00753_0053_000000000-D4HKJ is selected


slazicoicr:
Selecting 180810_M00753_0053_000000000-D4HKJ causes an exception:

```
Traceback (most recent call last):
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/slazic/.local/miniconda3/envs/work3.6/lib/python3.6/site-packages/dash/dash.py", line 1152, in dispatch
    response.set_data(self.callback_map[output]['callback'](*args))
[...]

PR #16, opened 2019-06-18
Fixed Bug #4

ijafri26:
Issue #4 
Created fresh branch for this pull request to prevent confusion with previous commits

### DIFF
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -93,6 +93,7 @@ def update_known_index_bar(run_json):
     run['library'] = run['SampleID'].str.extract(
         'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_'
     )
+    run['Index1'] = run['Index1'].astype('str')
     run['index'] = run['Index1'].str.cat(
         run['Index2'].fillna(''), sep=' '
     )


bcl2fastq/index_summary.py
```diff
@@ -93,6 +93,7 @@ def update_known_index_bar(run_json):
     run['library'] = run['SampleID'].str.extract(
         'SWID_\d+_(\w+_\d+_.*_\d+_[A-Z]{2})_'
     )
+    run['Index1'] = run['Index1'].astype('str')
```
slazicoicr: As I read it, this casting is done to convert `NaN` into the string `nan` and to allow the concatenation to occur in the next line.

I am not sure this is the best approach, as `nan` can be seen as a valid (if degenerate) DNA sequence. Replacing the `NaN` with an empty string would be the clearer approach.

callunity: To confirm -- the filtering should happen right after the assignment to `index` happens?

slazicoicr: Correct. 

This is a global variable shared by all connections to Dashi, so it makes sense to do sanity checks at this stage and avoid recalculations for each connection.

bcl2fastq/index_summary.py
```diff
@@ -5,11 +5,14 @@
...
 import pandas
 
-
 index = gsiqcetl.bcl2fastq.parse.load_cache(
     gsiqcetl.bcl2fastq.parse.CACHENAME.SAMPLES,
     './data/bcl2fastq_cache.hd5'
 )
+# If both Index1 and Index2 are NaN, this will filter out those indexes. This is to prevent run
+# 180810_M00753_0053_000000000-D4HKJ from crashing 
+if type(index[index['Index1'].notna()]) == float and type(index[index['Index2'].nota()]) == float:
```
slazicoicr: What is the purpose of this line? `type(index[index['Index1'].notna()]) == float` will always be False, as `type(index[index['Index1'].notna()])` is `DataFrame`.

bcl2fastq/index_summary.py
```diff
@@ -5,11 +5,14 @@
...
+    index = index[~index["Index1"].isna()]
```
slazicoicr: This only checks for Index1, but both Index1 and Index2 need to be NaN.

bcl2fastq/index_summary.py
ijafri26: My mistake, this should be .isna(). If I make this change, would that work?

slazicoicr: I don't believe so. What is the purpose of this if statement?

ijafri26: This was to filter by NaN, but I think I read the description wrong. NaN doesn't have type float so this statement wouldn't execute as I wanted.

slazicoicr: NaN is of type float

image

The line below (`index = index[~index["Index1"].isna()]`) does the filtering, so I am still not sure what the `if` statement aims to do.

ijafri26: It would only execute the filtering if both Index1 and Index 2 were float

slazicoicr: I understand. The `if` statement is not necessary. Once you remove 

https://github.com/ijafri26/dashi/blob/b73ad578089106ab89ecfccafaefc3282c71f824/bcl2fastq/index_summary.py#L99

you will see that the bug persists with your current code.

I think there is a misunderstanding in how DataFrame filtering works. I would suggest reading https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html and https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html

bcl2fastq/index_summary.py
```diff
@@ -4,15 +4,16 @@
 import dash_html_components as html
 import dash.dependencies as dep
 import pandas
+import math
 
-index = gsiqcetl.bcl2fastq.parse.load_cache(
+index1 = gsiqcetl.bcl2fastq.parse.load_cache(
```
callunity: I wouldn't rename this to `index1` as "Index1" already has a very specific meaning in the rest of the script ("the label for the Index 1 column") that doesn't correspond to this variable, which holds data for all the runs.
  (`index` is already an ambiguous name because of this, but changing it is beyond the scope of this issue)
Does the variable need to be renamed for line 16 to work, or can you keep using the same pattern as elsewhere in the file, like `index = index.dropna(...)`?

ijafri26: This variable needed to be renamed for a downstream function.  Line 177 requires a pure version of the variable to correctly display the pie graph. But I will rename it to something not already in use

bcl2fastq/index_summary.py
```diff
@@ -169,21 +174,21 @@ def update_pie_chart(run_alias, known_json, unknown_json):
     pruned_count = pruned['Count'].sum()
 
     total_clusters = gsiqcetl.bcl2fastq.utility.total_clusters_for_run(
-        run_alias, index
+        run_alias, index1
```
slazicoicr: Why would this function still require access the pre-filtered DataFrame? The run_alias options would not include any invalid runs, so the DataFrame does not need to contain them.

ijafri26: Without this change, it results in an error 
image

slazicoicr: That is a failed sanity check from GSI-QC-ETL. Let me investigate.

slazicoicr: Ahh, my past self saves me.

Currently, there are these possibilities for each lane of bcl2fastq analysis:

|Scenario| Index1 | Index2 |SampleID|Is it valid?|
| ---| --- | --- | --- | --- |
|1|ATGC|ATGC|Sample_1|Yes, dual index
|2|ATGC|NaN|Sample_1|Yes, single first index
|3|NaN|ATGC|Sample_1|Yes, dual second index
|4|NaN|NaN|NaN|Yes, the unknown indices in the lane
|5|NaN|NaN|Sample_1|No, a sample must have at least one index

Only Scenario 5 is invalid. With `index1.dropna(subset=('Index1','Index2'), thresh = 1)` we are getting rid of Scenario 4 and 5. When the total clusters are being calculated with `gsiqcetl.bcl2fastq.utility.total_clusters_for_run`, the unknown indices have been removed and the function correctly points out that those indices are needed to calculate the total clusters.

Irene's solution to this is to keep two DataFrames, which works. The issue is that now there are two very similar global DataFrames, making one more likely to use the wrong one and increasing memory footprint.

I can think of two solutions:

### Solution 1

Selectively remove runs that have at least 1 invalid library (while not removing Scenario 4 rows).  This requires a more challenging filter. It will require the use of `groupby` and `filter` functions (https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#).

These are the more complex and useful functions. To give you a hint, here is a way to filter for runs that have at least one library (row) in Scenario 1 (dual indexed):

`index.groupby('Run').filter(lambda x: any(~x['Index1'].isna() & ~x['Index2'].isna()))`

### Solution 2
Covert any instance of Scenario 5 into Scenario 4. In other words, if a row is invalid, it is treated as though this is an unknown index. This should be simpler to implement, but might require additional logic to deal with plotting runs that have no known indices.

Issue #3, opened 2019-06-06
bcl2fastq.index_summary: Select run from URL request


slazicoicr:
Currently, the latest run is selected by default. Add feature that allows the URL to specify which run to display on page load.

I think the easiest way is to use GET request. For example `www.oicrdashfake.com?run=190531_XAFAFSFE` should automatically select run 190531_XAFAFSFE

[Dash Location](https://dash.plot.ly/dash-core-components/location) should be used to implement this feature. Usage examples at https://dash.plot.ly/urls

PR #19, opened 2019-06-20
Resolution of Issue #3 Select Run from URL request

ijafri26:
Issue #3 
Created option to select or enter Run in URL  request to update page

### DIFF
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -115,10 +115,8 @@ def change_url(value):
             Error pop-up displayed depending on user input.
     """
     #   In a pathname, it automatically adds '/' to the beginning of the input even if pathname blank
-    if value == "/":
-        return all_runs[0], False
     #   While page loads, pathname is set to 'None'. Once page is loaded pathname is set to user input.
-    elif value is None:
+    if value == "/" or value is None:
         return all_runs[0], False
     elif value[1:] not in all_runs:
         return all_runs[0], True
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -48,10 +48,9 @@ layout = html.Div(children=[
     ),
     dcc.Dropdown(
         id='run_select',
-        #   Options and Value attributes are concantenated string versions of all_runs.
-        #   Value selects a single copy of the run name.
+        #   Options is concantenated string versions of all_runs.
         options=[{'label': r, 'value': r} for r in all_runs],
-        value=([{'label': r, 'value': r} for r in all_runs])[0]['value'],
+        value=all_runs[0],
         clearable=False
     ),
     dcc.Location(
@@ -115,15 +114,16 @@ def change_url(value):
             The string value (without '/') of the user input for the drop-down to use
             Error pop-up displayed depending on user input.
     """
-    #   In a pathname, it automatically adds '/' to the beginning of the input. The values in the drop-down do not have '/'
-    #   preceding the input value. Must convert to string and remove '/' to prevent errors
-
+    #   In a pathname, it automatically adds '/' to the beginning of the input even if pathname blank
     if value == "/":
-        return str(all_runs[0]), False
-    elif str(value)[1:] not in all_runs:
-        return str(all_runs[0]), True
+        return all_runs[0], False
+    #   While page loads, pathname is set to 'None'. Once page is loaded pathname is set to user input.
+    elif value is None:
+        return all_runs[0], False
+    elif value[1:] not in all_runs:
+        return all_runs[0], True
     else:
-        return str(value)[1:], False
+        return value[1:], False
 
 
 @app.callback(
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -15,10 +15,9 @@ unknown = gsiqcetl.bcl2fastq.parse.load_cache(
     './data/bcl2fastq_cache.hd5'
 )
 
-all_runs = index['Run'].sort_values(ascending=False).unique()
-
-all_runs = [{'label': r, 'value': r} for r in all_runs]
-
+all_runsog = index['Run'].sort_values(ascending=False).unique()
+print (all_runsog)
+all_runs = [{'label': r, 'value': r} for r in all_runsog]
 """ Sample_run_hidden holds json split format of "Known"
         columns: "FlowCell","Index1","Index2","LIMS IUS SWID","LaneClusterPF","LaneClusterRaw",
                  "LaneNumber","LaneYield","QualityScoreSum","ReadNumber","Run","RunNumber","SampleID",
@@ -45,6 +44,10 @@ all_runs = [{'label': r, 'value': r} for r in all_runs]
                 -  multiple select gives error due to inconsistent values
           """
 layout = html.Div(children=[
+    dcc.ConfirmDialog(
+        id='error',
+        message='You have input an incorrect run. Click "Ok" to return to the most recent run.'
+    ),
     dcc.Dropdown(
         id='run_select',
         options=all_runs,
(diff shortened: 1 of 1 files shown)


slazicoicr:
Three points:

### 1
Does not seem to work. When I enter `http://127.0.0.1:8050/190530_A00469_0041_AHHMTVDSXX` the page refreshes to `http://127.0.0.1:8050/190606_M00146_0057_000000000-D6D2L`

### 2
The issue specifies the use of a GET request, such as `http://127.0.0.1:8050?run=190530_A00469_0041_AHHMTVDSXX`. This allow for future expansion if additional data needs to be specified.

### 3
The lines

https://github.com/oicr-gsi/dashi/blob/609b2f7d8931c0e86d9f04b226e8f14edcfbc528/bcl2fastq/index_summary.py#L190-L199

are changed the same way in this pull request and the merged #13 . This shouldn't be happening.

callunity:
To Savo's three points:
1. Yes, this doesn't seem to be updating the graphs with data for the requested run.
2. @slazicoicr : will this page ever display more than one run, or no runs? (If it's only ever going to display one run, it makes semantic sense to set the run name in the path and it could be extended by requesting additional data by query parameters)
3. Looks to be fixed by the rebase

slazicoicr:
Regarding point 2. This visualization was designed with one run in mind, so Irene can pick the URL strategy that is the cleanest to implement.

bcl2fastq/index_summary.py
```diff
@@ -15,10 +15,9 @@
     './data/bcl2fastq_cache.hd5'
 )
 
-all_runs = index['Run'].sort_values(ascending=False).unique()
-
-all_runs = [{'label': r, 'value': r} for r in all_runs]
-
+all_runsog = index['Run'].sort_values(ascending=False).unique()
+print (all_runsog)
```
slazicoicr: Stray print statement. I suggest using the `git diff` command to check the changes you have introduced before comitting them. This command is also a [PyCharm feature](https://www.jetbrains.com/help/pycharm/comparing-files-and-folders.html)

bcl2fastq/index_summary.py
```diff
@@ -15,10 +15,9 @@
...
+all_runs = [{'label': r, 'value': r} for r in all_runsog]
```
slazicoicr: Is it possible to avoid having two global variables (`all runs` and `all_runsog`) with very similar names and content? If not, please use clearer names and comments explaining the purpose of each. Global variables can be used anywhere, so their names should be very clear.

bcl2fastq/index_summary.py
```diff
@@ -45,6 +44,10 @@
                 -  multiple select gives error due to inconsistent values
           """
 layout = html.Div(children=[
+    dcc.ConfirmDialog(
+        id='error',
+        message='You have input an incorrect run. Click "Ok" to return to the most recent run.'
```
slazicoicr: What happens when you click `Cancel`?

bcl2fastq/index_summary.py
```diff
@@ -100,13 +100,19 @@
...
-    [dep.Input('run_select', 'value')]
+    [dep.Output('run_select', 'value'),
+     dep.Output('error', 'displayed')],
+    [dep.Input('url', 'pathname')]
 )
-def change_url(dropdown_value):
-    """Allows user to enter Run name in URL which will update dropdown automatically, and the graphs
+def change_url(value):
+    """Allows user to enter Run name in URL which will update dropdown automatically, and the graphs.
+    If User enters any value that's not a valid run an error box will pop up and return user to most recent run
```
slazicoicr: Specify Parameters and Return

bcl2fastq/index_summary.py
```diff
@@ -100,13 +100,19 @@
...
     """
-    return dropdown_value
+
+    if value == "/" or str(value)[1:] not in all_runsog:
```
slazicoicr: Why the `str` conversion? The `pathname` variable is a string: https://dash.plot.ly/dash-core-components/location. Same question for the other uses of `str` in this function.

slazicoicr: Add a comment explaining why you are getting everything but the first character of a string.

bcl2fastq/index_summary.py
```diff
@@ -153,7 +159,8 @@ def update_pruned_unknown_hidden(run_alias):
 
 @app.callback(
     dep.Output('known_index_bar', 'figure'),
-    [dep.Input('sample_run_hidden', 'children')]
+    [dep.Input('sample_run_hidden', 'children')
+     ]
```
slazicoicr: Inconsistent indentation.

callunity: Please reformat the code before each commit. In PyCharm this is done with Code > Reformat Code, or `Ctrl+Alt+L`. You can also install the git pre-commit hook which will do it for you.

bcl2fastq/index_summary.py
ijafri26: Nothing, it does the same thing as if you click 'ok'. There doesn't seem to be a way to only show one button

slazicoicr: Clarify that in the Message string, so the user knows it doesn't matter which button they click.

bcl2fastq/index_summary.py
ijafri26: If we don't convert it, it gives the following error
image

bcl2fastq/index_summary.py
```diff
@@ -44,18 +42,27 @@
                 -  multiple select gives error due to inconsistent values
           """
 layout = html.Div(children=[
+    dcc.ConfirmDialog(
+        id='error',
+        message='You have input an incorrect run. Click either "Ok" or "Cancel" to return to the most recent run.'
```
callunity: Since clicking either button brings the user back to the same place, it would make more sense for the user if we use an "OK" dialog instead of a "Confirm" dialog (with updated message text to match).

bcl2fastq/index_summary.py
```diff
@@ -44,18 +42,27 @@
...
+    ),
     dcc.Dropdown(
         id='run_select',
-        options=all_runs,
-        value=all_runs[0]['value'],
+        #   Options and Value attributes are concantenated string versions of all_runs.
+        #   Value selects a single copy of the run name.
+        options=[{'label': r, 'value': r} for r in all_runs],
+        value=([{'label': r, 'value': r} for r in all_runs])[0]['value'],
```
callunity: I think this can be simplified -- `all_runs` is a list of run names, and we want the first item in this list to be selected by default. `value` is the string value of the item to be selected. Would it work as `value=all_runs[0]`?

bcl2fastq/index_summary.py
callunity: I think that happens in the case where you load the page and there's no text after the port (ie there's no "/"). Can this if statement be changed to handle that case as well? And then the comment could explain that it starts counting after the initial "/".

bcl2fastq/index_summary.py
ijafri26: Yes it works, updating commit

bcl2fastq/index_summary.py
slazicoicr: You are converting `None` (https://stackoverflow.com/questions/19473185/what-is-a-none-value), which is of type `NoneType` to `'None'`, which is a `str` type. 

This approach is a bug. `None` means there is no data and by converting it to the string `'None'` you are creating invalid data.

Really engage with the exception before attempting a fix. The `TypeError` exception states that `None[1:]` cannot be done. Your solution is to do `'None'[1:]`, which does work, but makes no sense. Since `'None'[1:] not in all_runsog` will always be true (unless we have a run named `one`), everything seems to work, but the underlying problem is not solved.

Issue #12, opened 2019-06-11
bcl2fastq.index_summary: Wrong counting in Samples Indices graph


slazicoicr:
The number of reads per index are counted here

https://github.com/oicr-gsi/dashi/blob/a77647182ef0356d22871f79ae5bf0552378ac14/bcl2fastq/index_summary.py#L101-L103

The assumption is that an index corresponds to a library, which is false. Within one run, the same index can be used for more than one library by putting different libraries with the same index on different lanes.

This occurs in 190531_A00469_0042_BHJJG2DSXX, where CPTP_0163_Pb_P_PE_491_SC and CPTP_0309_Pb_P_PE_494_SC share the same index, but are on different lanes.

The groupby statement needs to take into account both index and library.

PR #22, opened 2019-06-25
Issue #12 Wrong counting in Samples Indices graph

ijafri26:
Issue #12 
Grouped data for Samples Indices graph (line 194) by both index and library to prevent incorrect counting of data.

### DIFF
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -193,9 +193,9 @@ def update_known_index_bar(run_json):
     #   to prevent duplicate counts
     for inx, d in run.groupby(['index', 'library']):
         data.append({
-            'x': list(d['library']),
+            'x': list(d['library'].unique()),
             # One library can be run on multiple lanes. Sum them together.
-            'y': list(d.groupby('library')['SampleNumberReads'].sum()),
+            'y': [d['SampleNumberReads'].sum()],
             'type': 'bar',
             'name': inx[0],
             'marker': {'line': {
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -189,13 +189,15 @@ def update_known_index_bar(run_json):
     )
 
     data = []
-    for inx, d in run.groupby('index'):
+    #   Multiple libraries can use the same index, data must be grouped by both index and libraries
+    #   to prevent duplicate counts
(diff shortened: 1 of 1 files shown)


bcl2fastq/index_summary.py
```diff
@@ -189,13 +189,15 @@ def update_known_index_bar(run_json):
...
 
     data = []
-    for inx, d in run.groupby('index'):
+    #   Multiple libraries can use the same index, data must be grouped by both index and libraries
+    #   to prevent duplicate counts
+    for inx, d in run.groupby(['index', 'library']):
         data.append({
             'x': list(d['library']),
             # One library can be run on multiple lanes. Sum them together.
             'y': list(d.groupby('library')['SampleNumberReads'].sum()),
```
slazicoicr: Since you already grouped by library, there is no need to group again.

callunity:
> Use the [re-request review function](https://help.github.com/en/articles/requesting-a-pull-request-review) (Point # 7) to indicate that your commits are ready for another review.

@slazicoicr am I missing something? It doesn't look like the commits had been reviewed a first time, or that you had been previous requested for a review, so I'm not clear on why she'd need to request a re-review...

slazicoicr:
My understanding of this feature was that since I asked for changes in this pull request, Irene could commit several times. I would not know when she is ready for this Pull Request to be reviewed again. By explicitly pressing the re-review button on my name, she would be stating that I should review the Pull Request again.

callunity:
Ah, I see; I thought you were commenting on her previous use of the review request feature, as opposed to requesting a specific use pattern going forward. This makes more sense.

slazicoicr:
I think this approach fixes the bug. @ijafri26 let us know how this differs from your approach that failed to produce valid outputs:

```
     data = []
-    for inx, d in run.groupby('index'):
+    for inx, d in run.groupby(['index', 'library']):
         data.append({
-            'x': list(d['library']),
+            'x': list(d['library'].unique()),
             # One library can be run on multiple lanes. Sum them together.
-            'y': list(d.groupby('library')['SampleNumberReads'].sum()),
+            'y': [d['SampleNumberReads'].sum()],
             'type': 'bar',
-            'name': inx,
+            'name': inx[0],
             'marker': {'line': {
                 'width': 2,
                 'color': 'rgb(255,255, 255)'
```

Issue #9, opened 2019-06-07
bcl2fastq.index_summary: Remove hidden Divs


slazicoicr:
This visualization was started before realizing that multiple Outputs were implemented in Dash: https://github.com/plotly/dash/pull/436. This led to usage of hidden Divs (to get around lack of multiple Outputs) and multiple Outputs (once realizing that they were implemented).

Remove hidden Divs and replace with multiple Outputs design. 

Hidden Divs are used

https://github.com/oicr-gsi/dashi/blob/a77647182ef0356d22871f79ae5bf0552378ac14/bcl2fastq/index_summary.py#L51-L52

and multiple Outputs are used

https://github.com/oicr-gsi/dashi/blob/a77647182ef0356d22871f79ae5bf0552378ac14/bcl2fastq/index_summary.py#L157-L162

PR #23, opened 2019-06-27
Issue#9: Removed hidden Divs and combined Outputs

ijafri26:
FIxed Issue #9 and removed hidden divs. Created one app.Callback (Update Layout) that has multiple outputs to consolidate all the other functions

### DIFF
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -138,7 +138,7 @@ def update_known_index_bar(run_alias):
 
 
 def update_unknown_index_bar(run_alias):
-    """ Function to unknown index bar  according to user selected run
+    """ Function to update unknown index bar  according to user selected run
             Parameters:
                  run_alias: name of run
             Returns:
@@ -181,16 +181,16 @@ def update_pie_chart(run_alias):
                  updated pie chart "known_unknown_pie" with known and unknown indices ratio over total cluster
                  updates value of known_fraction
      """
-    known = index[index['Run'] == run_alias]
-    known = known[known['ReadNumber'] == 1]
-    known = known[~known['SampleID'].isna()]
-    known = known.drop_duplicates(['SampleID', 'LaneNumber'])
+    run = index[index['Run'] == run_alias]
+    run = run[run['ReadNumber'] == 1]
+    run = run[~run['SampleID'].isna()]
+    run = run.drop_duplicates(['SampleID', 'LaneNumber'])
 
     pruned = gsiqcetl.bcl2fastq.utility.prune_unknown_index_from_run(
         run_alias, index, unknown
     )
 
-    known_count = known['SampleNumberReads'].sum()
+    known_count = run['SampleNumberReads'].sum()
     pruned_count = pruned['Count'].sum()
     total_clusters = gsiqcetl.bcl2fastq.utility.total_clusters_for_run(
         run_alias, index
@@ -216,12 +216,12 @@ def update_pie_chart(run_alias):
      dep.Output('error', 'displayed')],
     [dep.Input('url', 'pathname')]
 )
-def change_url(value):
+def change_url(pathname):
     """ Allows user to enter Run name in URL which will update dropdown automatically, and the graphs.
         If User enters any value that's not a valid run an error box will pop up and return user to most recent run
 
         Parameters:
-             value: the URL input after the port (the pathname).
+             pathname: user-requested path.
 
         Returns:
             The string value (without '/') of the user input for the drop-down to use
@@ -229,12 +229,12 @@ def change_url(value):
     """
     #   In a pathname, it automatically adds '/' to the beginning of the input even if pathname blank
     #   While page loads, pathname is set to 'None'. Once page is loaded pathname is set to user input.
-    if value == "/" or value is None:
+    if pathname == "/" or pathname is None:
         return all_runs[0], False
-    elif value[1:] not in all_runs:
+    elif pathname[1:] not in all_runs:
         return all_runs[0], True
     else:
-        return value[1:], False
+        return pathname[1:], False
 
 
 @app.callback(
@@ -243,16 +243,16 @@ def change_url(value):
      dep.Output('known_unknown_pie', 'figure'),
      dep.Output('known_fraction', 'value')],
     [dep.Input('run_select', 'value')]
-
 )
 def update_layout(run_alias):
-    """ When input(run dropdown) is changed, known index bar, unknown index bar, piechart and textarea are called
+    """ When input(run dropdown) is changed, known index bar, unknown index bar, piechart and textarea are updated
         Returns:
             functions update_known_index_bar, update_unknown_index_bar, update_pie_chart's data value,
             and update_pie_chart's fraction value
     """
-    return update_known_index_bar(run_alias), update_unknown_index_bar(run_alias), update_pie_chart(run_alias)[0], \
-           update_pie_chart(run_alias)[1]
+    pie_data, textarea_fraction = update_pie_chart(run_alias)
+
+    return update_known_index_bar(run_alias), update_unknown_index_bar(run_alias), pie_data, textarea_fraction
 
 
 if __name__ == '__main__':
--- a/bcl2fastq/index_summary.py
+++ b/bcl2fastq/index_summary.py
@@ -20,11 +20,11 @@ all_runs = index['Run'].sort_values(ascending=False).unique()
         columns: "FlowCell","Index1","Index2","LIMS IUS SWID","LaneClusterPF","LaneClusterRaw",
                  "LaneNumber","LaneYield","QualityScoreSum","ReadNumber","Run","RunNumber","SampleID",
                  "SampleName","SampleNumberReads","SampleYield","TrimmedBases","Yield","YieldQ30"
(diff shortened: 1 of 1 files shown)


bcl2fastq/index_summary.py
```diff
@@ -201,50 +127,43 @@ def update_known_index_bar(run_json):
...
-                   run_json(str): json format of run data
-
-               Returns:
-                   data and layout values for stacked bar graph for unknown indices
-                   updates unknown_index_bar bar graph
-           """
-    # pruned(str): holds first 30 unknown indices
-    pruned = pandas.read_json(run_json, orient='split')
+def update_unknown_index_bar(run_alias):
+    """ Function to unknown index bar  according to user selected run
```
callunity: `Function to unknown index bar ...` -> `Function to update unknown index bar graph ...` ?

bcl2fastq/index_summary.py
```diff
@@ -254,34 +173,29 @@ def update_unknown_index_bar(run_json):
...
+             Parameters:
+                 run_alias(str): name of run
+             Returns:
+                 updated pie chart "known_unknown_pie" with known and unknown indices ratio over total cluster
+                 updates value of known_fraction
+     """
+    known = index[index['Run'] == run_alias]
+    known = known[known['ReadNumber'] == 1]
+    known = known[~known['SampleID'].isna()]
+    known = known.drop_duplicates(['SampleID', 'LaneNumber'])
```
callunity: `known` makes me think that this variable should hold something about the known indices, but lines 107-110 above have the exact same functionality, but calls the variable `run`. Can this be standardized to prevent confusion?

bcl2fastq/index_summary.py
```diff
@@ -297,5 +211,49 @@ def update_pie_chart(run_alias, known_json, unknown_json):
...
+    [dep.Output('run_select', 'value'),
+     dep.Output('error', 'displayed')],
+    [dep.Input('url', 'pathname')]
+)
+def change_url(value):
+    """ Allows user to enter Run name in URL which will update dropdown automatically, and the graphs.
+        If User enters any value that's not a valid run an error box will pop up and return user to most recent run
+
+        Parameters:
+             value: the URL input after the port (the pathname).
```
callunity: `the URL input after the port` -> `the user-requested path` ? (We could host the website so the port isn't displayed in the URL, but this function would still work.
A quick note about naming: it's preferable to have more specific names (for instance, `path` rather than `value`). This helps the reader since they don't have to remember what the more general name refers to.

bcl2fastq/index_summary.py
```diff
@@ -297,5 +211,49 @@ def update_pie_chart(run_alias, known_json, unknown_json):
...
+@app.callback(
+    [dep.Output('known_index_bar', 'figure'),
+     dep.Output('unknown_index_bar', 'figure'),
+     dep.Output('known_unknown_pie', 'figure'),
+     dep.Output('known_fraction', 'value')],
+    [dep.Input('run_select', 'value')]
+
+)
+def update_layout(run_alias):
+    """ When input(run dropdown) is changed, known index bar, unknown index bar, piechart and textarea are called
```
callunity: `are called` -> `are updated`?

bcl2fastq/index_summary.py
```diff
@@ -297,5 +211,49 @@ def update_pie_chart(run_alias, known_json, unknown_json):
...
+        Returns:
+            functions update_known_index_bar, update_unknown_index_bar, update_pie_chart's data value,
+            and update_pie_chart's fraction value
+    """
+    return update_known_index_bar(run_alias), update_unknown_index_bar(run_alias), update_pie_chart(run_alias)[0], \
```
callunity: Instead of calling `update_pie_chart` twice to get the two values, you could call it once and use multiple variable assignment to save the resulting values, and then `return` those.

bcl2fastq/index_summary.py
```diff
@@ -98,99 +96,27 @@
...
-
-
-@app.callback(
-    dep.Output('known_index_bar', 'figure'),
-    [dep.Input('sample_run_hidden', 'children')
-     ]
-)
-def update_known_index_bar(run_json):
-    """ When input (sample_run_hidden) is changed, function is rerun to get update bar graph of sample indices
+def update_known_index_bar(run_alias):
```
slazicoicr: These functions no longer update the layout graphs, but create the graphs (this applies to all non-callback functions). Maybe `create_known_index_bar`.

bcl2fastq/index_summary.py
```diff
@@ -98,99 +96,27 @@
...
+               run_alias: name of run
            Returns:
               data and values for the layout of stacked bar graph of sample indices
               updates bar graph "known_index_bar"
        """
-    run = pandas.read_json(run_json, orient='split')
+    run = index[index['Run'] == run_alias]
+    run = run[run['ReadNumber'] == 1]
+    run = run[~run['SampleID'].isna()]
+    run = run.drop_duplicates(['SampleID', 'LaneNumber'])
```
slazicoicr: Calculating the `run` DataFrame is done twice (the other time is on Line 184). Move the calculation into `update_layout` and pass the DataFrame into this function.

bcl2fastq/index_summary.py
```diff
@@ -201,50 +127,43 @@ def update_known_index_bar(run_json):
...
+    """ Function to update unknown index bar  according to user selected run
+            Parameters:
+                 run_alias: name of run
+            Returns:
+                data and layout values for stacked bar graph for unknown indices
+                updates unknown_index_bar bar graph
+              """
+    pruned = gsiqcetl.bcl2fastq.utility.prune_unknown_index_from_run(
+        run_alias, index, unknown
+    )
```
slazicoicr: Creating the `pruned` DataFrame is done twice (on line 189). Move the calculation into `update_layout` and pass the DataFrame.

bcl2fastq/index_summary.py
```diff
@@ -297,5 +211,49 @@ def update_pie_chart(run_alias, known_json, unknown_json):
     )
```
slazicoicr: It is very hard to visually see that this return statement gives back two values. Reformat it with newlines/spacing so it is more obvious.

bcl2fastq/index_summary.py
```diff
@@ -98,99 +96,17 @@
...
-     ]
-)
-def update_known_index_bar(run_json):
-    """ When input (sample_run_hidden) is changed, function is rerun to get update bar graph of sample indices
+def create_known_index_bar(run):
+    """ Function to update known index bar according to user selected run
            Parameters:
-               run_json(str): json format of run data
-
+               run_alias: name of run
```
slazicoicr: Documentation no longer matches function call.

bcl2fastq/index_summary.py
```diff
@@ -201,50 +117,35 @@ def update_known_index_bar(run_json):
...
-    pruned = pandas.read_json(run_json, orient='split')
-    pruned['index'] = pruned['Index1'].str.cat(
-        pruned['Index2'].fillna(''), sep=' '
-    )
-    pruned = pruned.sort_values('Count', ascending=False)
-    pruned = pruned.head(30)
+def create_unknown_index_bar(pruned):
+    """ Function to update unknown index bar  according to user selected run
+            Parameters:
+                 run_alias: name of run
```
slazicoicr: Documentation needs to be updated.

bcl2fastq/index_summary.py
```diff
@@ -254,47 +155,99 @@ def update_unknown_index_bar(run_json):
     }
 
 
+def create_pie_chart(run, pruned, total_clusters):
+    """ Function to update pie chart and known fraction according to user selected run
+             Parameters:
+                 run_alias(str): name of run
```
slazicoicr: Update parameter documentation.

bcl2fastq/index_summary.py
```diff
@@ -254,47 +155,99 @@ def update_unknown_index_bar(run_json):
...
+@app.callback(
+    [dep.Output('known_index_bar', 'figure'),
+     dep.Output('unknown_index_bar', 'figure'),
+     dep.Output('known_unknown_pie', 'figure'),
+     dep.Output('known_fraction', 'value')],
+    [dep.Input('run_select', 'value')]
+)
+def update_layout(run_alias):
+    """ When input(run dropdown) is changed, known index bar, unknown index bar, piechart and textarea are updated
+        Returns:
```
slazicoicr: Parameter documentation missing.

slazicoicr:
@ijafri26 There are merge conflicts with the Pull Requests resulting from fixing Issue #12 .

bcl2fastq/index_summary.py
```diff
@@ -256,47 +157,103 @@ def update_unknown_index_bar(run_json):
...
+    return {'data': [{
+        'labels': ['Known', 'Unknown'],
+        'values': [known_count, pruned_count],
+        'type': 'pie',
+        'marker': {'colors': ['#349600', '#ef963b']},
+                      }],
+            'layout': {
+              'title': 'Flow Cell Composition of Known/Unknown Indices'}
+
+           }, \
```
slazicoicr: I don't think `\` should be necessary here.

PR #24, opened 2019-07-15
Pooling qc

ijafri26:
Created PoolingQC report sample for JIRA tickets:
GR-755 -  Allow user to request a pool balancing report by URL
https://jira.oicr.on.ca/browse/GR-755
GR-756 - Create a page got the pool balancing report
https://jira.oicr.on.ca/browse/GR-756
GR-757 - Display thresholds and highlight failed samples/metrics
https://jira.oicr.on.ca/browse/GR-757
GR-762 - Investigate how to change the styling on certain cells in a table or items in a graph
https://jira.oicr.on.ca/browse/GR-762


index.py
```diff
@@ -10,10 +10,11 @@
 from rnaseqc.over_time import layout as rnaseqc_overtime_layout
 from runreport.proj_hist import layout as runreport_projhist_layout
 from runscanner.yield_over_time import layout as runscanner_yield_over_time_layout
+from poolingQC_sample import layout as poolingQC_layout
```
callunity: Could `poolingQC_sample` please be converted to `pooling_qc_sample` to match the snake case throughout the rest of the project?

index.py
```diff
@@ -33,6 +34,8 @@ def display_page(pathname):
         return runreport_projhist_layout
     elif pathname == '/runscanner/sum_over_time':
         return runscanner_yield_over_time_layout
+    elif pathname == '/poolingQC':
```
callunity: Please also update this to snake case (`pooling_qc`).

poolqc/poolingQC_sample.py
```diff
@@ -0,0 +1,300 @@
...
+import dash_html_components as html
+import dash.dependencies as dep
+import dash_table as dt
+import pandas
+import plotly.plotly as plotly
+import plotly.graph_objs as go
+import numpy as np
+
+pandas.options.mode.chained_assignment = None
+index = pandas.read_hdf('/home/ijafari/cache_files/rnaseqqc_cache.hd5')
```
callunity: Please don't hard-code paths here, as nobody else's computer will have a `/home/ijafari` folder on it! You should be able to use relative paths as you know the rest of the dashi project expects to find cache files in a `data` folder located at the root of the dashi project.

poolqc/poolingQC_sample.py
```diff
@@ -0,0 +1,300 @@
...
+
+all_runs = index['Sequencer Run Name'].sort_values(ascending=False).unique()
+all_runs = [x for x in all_runs if str(x) != 'nan']
```
callunity: Can `dropna` or `notnull` be used in place of `!= 'nan'`?

poolqc/poolingQC_sample.py
```diff
@@ -0,0 +1,300 @@
...
+
+layout = html.Div(children=[
+    dcc.ConfirmDialog(
+        id='error',
+        message='You have input an incorrect URL. Click either "Ok" or "Cancel" to return to the most run.'
```
callunity: Can this error message be more informative than "incorrect URL"? I believe that no data would be found for the given run if either the run is misspelled or analysis has not yet completed for this run.

poolqc/poolingQC_sample.py
apmasell: Or use `argparse` so the directories can be specified on the command line.

poolqc/poolingQC_sample.py
ijafri26: If the run is misspelled the report will automatically return to the most recent analysis completed.

"The analysis on this run has not been yet completed. Click either 'Ok' or 'Cancel to return to the most recent analysis completed."
Would this be a good alternative?

callunity: Are you able to distinguish between runs that have not completed analysis (say, those that have only gone through bcl2fastq) and runs that are misspelled?

poolqc/poolingQC_sample.py
callunity: @apmasell I don't know if that'll work when Dash is running in web server mode...

poolqc/poolingQC_sample.py
ijafri26: all_runs is a list and df.dropna or df.notnull doesn't seem to work on it

callunity: Ah, it looks like the sequencer run name column already contains the string literal `'nan'`. The cast to `str(x)` is not strictly necessary, but probably worth keeping in to future-proof it.

index.py
```diff
@@ -10,10 +10,11 @@
 from rnaseqc.over_time import layout as rnaseqc_overtime_layout
 from runreport.proj_hist import layout as runreport_projhist_layout
 from runscanner.yield_over_time import layout as runscanner_yield_over_time_layout
+from pooling_qc_sample import layout as pooling_qc_layout
```
callunity: `pooling_qc_sample` -> `poolqc.pooling_qc_sample` (`name_of_directory.name_of_file`)

poolqc/pooling_qc_sample.py
```diff
@@ -0,0 +1,408 @@
...
+import dash.dependencies as dep
+import dash_table as dt
+import pandas
+import plotly.plotly as plotly
+import plotly.graph_objs as go
+import numpy as np
+
+
+pandas.options.mode.chained_assignment = None
+index = pandas.read_hdf('../data/rnaseqqc_cache.hd5')
```
callunity: This should be `./data/rnaseqqc_cache.hd5` (note the single `.`), as per the other modules in this project. I've created ticket #25 which addresses this issue, but there's no need to add a fix to #25 to this PR.

index.py
callunity: Please update this.

poolqc/pooling_qc_sample.py
callunity: Please update this.

index.py
```diff
@@ -35,7 +35,7 @@ def display_page(pathname):
     elif pathname == '/runscanner/sum_over_time':
         return runscanner_yield_over_time_layout
     elif pathname == '/pooling_qc':
-        return pooling_qc_layout
+        return poolqc.pooling_qc_layout
```
callunity: Doing `import x as y` (line 13) means the function gets renamed to `y` for the duration of the file, so just `pooling_qc_layout`

poolqc/pooling_qc_sample.py
callunity: Please update this.

PR #33, opened 2019-08-01
Visualizations improvements - Removal of interactivity, and extra uneeded graphs

ijafri26:
Removed Interactivity from report as requested in the meeting. This will also prevent further complications with report

### DIFF
--- a/dashi/bcl2fastq/index_summary.py
+++ b/dashi/bcl2fastq/index_summary.py
@@ -54,7 +54,7 @@ layout = html.Div(children=[
         clearable=False
     ),
     dcc.Location(
-        id='url',
+        id='bcl2fastq_url',
         refresh=False
     ),
 
@@ -188,7 +188,7 @@ def create_pie_chart(run, pruned, total_clusters):
 @app.callback(
     [dep.Output('run_select', 'value'),
      dep.Output('error', 'displayed')],
-    [dep.Input('url', 'pathname')]
+    [dep.Input('bcl2fastq_url', 'pathname')]
 )
 def change_url(pathname):
     """ Allows user to enter Run name in URL which will update dropdown automatically, and the graphs.
@@ -205,10 +205,10 @@ def change_url(pathname):
     #   While page loads, pathname is set to 'None'. Once page is loaded pathname is set to user input.
     if pathname == "/" or pathname is None:
         return all_runs[0], False
-    elif pathname[1:] not in all_runs:
+    elif pathname[1:-2] not in all_runs:
         return all_runs[0], True
     else:
-        return pathname[1:], False
+        return pathname[1:-2], False
 
 
 @app.callback(
--- a/dashi/poolqc/pooling_qc_sample.py
+++ b/dashi/poolqc/pooling_qc_sample.py
@@ -1,10 +1,12 @@
 import gsiqcetl.bcl2fastq.parse
 import gsiqcetl.bcl2fastq.utility
 import dash_core_components as dcc
+import dash_table.FormatTemplate as FormatTemplate
 import dash_html_components as html
 import dash.dependencies as dep
 import dash_table as dt
 import pandas
+import plotly.plotly as plotly
 import plotly.graph_objs as go
 import numpy as np
 import urllib
@@ -24,6 +26,9 @@ bcl2fastq['library'] = bcl2fastq['SampleID'].str.extract('SWID_\d+_(\w+_\d+_.*_\
 
 df = bcl2fastq.merge(rnaseq, on='library', how='outer')
 df = df.merge(bamqc, on='library', how='outer')
+
+df['properly paired reads'] = np.where(df['properly paired reads'].isnull(),
+                                       df['rRNA Contamination (%reads properly paired)'], df['properly paired reads'])
 
 df = df.drop(columns=['Sequencer Run Name', 'Lane Number'])
 
@@ -54,6 +59,9 @@ layout = html.Div(children=[
         )),
     html.Div(id='Title', style={'fontSize': 25, 'textAlign': 'center', 'padding': 30}),
     html.Div(children=''),
+    html.Div(
+        dcc.Graph(id='SampleIndices'),
+    ),
     html.A(
         'Download Data',
         id='download-link',
@@ -80,14 +88,15 @@ layout = html.Div(children=[
             style_header={'backgroundColor': 'rgb(222,222,222)',
                           'fontSize': 16,
                           'fontWeight': 'bold'},
+            n_fixed_columns=2,
 
         )),
+
     html.Div(
-        dcc.Graph(id='SampleIndices'),
+        dcc.Graph(id='properly paired reads RNA', style={'display': 'none'}),
     ),
     html.Div(
-        dcc.Graph(id='map_code')
-    )
+        dcc.Graph(id='properly paired reads DNA', style={'display': 'none'}))
 ]
 )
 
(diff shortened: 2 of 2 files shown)


poolqc/pooling_qc_sample.py
```diff
@@ -197,45 +203,32 @@ def Summary_table(run_alias, lane_alias):
...
 
-def update_sampleindices(run, rows, derived_virtual_selected_rows, colors):
+@app.callback(
+    dep.Output('SampleIndices', 'figure'),
+    [dep.Input('select_a_run', 'value'),
+     dep.Input('lane_select', 'value')])
+def update_sampleindices(run_alias, lane_alias):
+    run = df[(df['Run'] == run_alias) & (df['LaneNumber'] == lane_alias)]
+    run = run[~run['Run'].isna()].drop_duplicates('library')
+    run['Result'] = run['Coding Bases'] / run['Passed Filter Aligned Bases'] * 100
```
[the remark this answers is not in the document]
callunity: If this is calculating % `Coding Bases` / `PF Aligned Bases`, then should the graph be renamed? I would expect a graph titled "Index Clusters" would be drawing from the `SampleNumberReads` metric.

poolqc/pooling_qc_sample.py
```diff
@@ -197,45 +203,32 @@ def Summary_table(run_alias, lane_alias):
...
+
+    total_RNA = len(run['library'])
+    Index_Pass = '%s/%s' % (sum(i > 20000 for i in run['SampleNumberReads']), total_RNA)
+    Index_Threshold = 50000000
     data = []
 
-    for inx, d in run.groupby(['index']):
+    for inx, d in run.groupby(['library']):
+        d['Threshold'] = Index_Threshold
+        d['Color'] = np.where((d['SampleNumberReads'] >= d['Threshold']), '#ffffff', '#db4b4b')
```
callunity: Can the assignment to `d['Threshold']` be avoided by using `np.where((d['SampleNumberReads'] >= Index_Threshold), ...)`?

poolqc/pooling_qc_sample.py
```diff
@@ -197,45 +203,32 @@ def Summary_table(run_alias, lane_alias):
...
+def update_sampleindices(run_alias, lane_alias):
+    run = df[(df['Run'] == run_alias) & (df['LaneNumber'] == lane_alias)]
+    run = run[~run['Run'].isna()].drop_duplicates('library')
+    run['Result'] = run['Coding Bases'] / run['Passed Filter Aligned Bases'] * 100
+    run['index'] = run['Index1'].str.cat(
+        run['Index2'].fillna(''), sep=' ')
+    run = run.sort_values('library')
+
+    total_RNA = len(run['library'])
+    Index_Pass = '%s/%s' % (sum(i > 20000 for i in run['SampleNumberReads']), total_RNA)
```
callunity: This might more accurately be called `samples_passing_clusters` as it's not describing the number of indices which passed.
Please use snake case (lowercase words separated by underscores) for your variable names.

poolqc/pooling_qc_sample.py
```diff
@@ -244,56 +237,26 @@ def update_percent_mapped_to_code(run, rows, derived_virtual_selected_rows, colo
...
 
 
-@app.callback(
-    [dep.Output('map_code', 'figure'),
-     dep.Output('SampleIndices', 'figure')],
-    [dep.Input('select_a_run', 'value'),
-     dep.Input('lane_select', 'value'),
-     dep.Input('Summary Table', 'derived_virtual_data'),
-     dep.Input('Summary Table', 'derived_virtual_selected_rows')])
-def update_all_graphs(run_alias, lane_alias, rows, derived_virtual_selected_rows):
```
callunity: I'd prefer to keep an "update all graphs" function which wraps `update_sampleindices`, as it'll make it easier if we need to add new graphs later.

poolqc/pooling_qc_sample.py
```diff
@@ -27,6 +25,9 @@
 df = bcl2fastq.merge(rnaseq, on='library', how='outer')
 df = df.merge(bamqc, on='library', how='outer')
 
+df['properly paired reads'] = np.where(df['properly paired reads'].isnull(),
+                                       df['rRNA Contamination (%reads properly paired)'], df['properly paired reads'])
```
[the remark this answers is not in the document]
ijafri26: This should be removed, I discussed with Savo earlier on. Will update

poolqc/pooling_qc_sample.py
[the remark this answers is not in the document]
ijafri26: This was accidentally left in prior to the removal of the extra graphs.  I have removed it in an upcoming commit

poolqc/pooling_qc_sample.py
ijafri26: The assignment was done for the threshold graph (`go.scatter`) as the values must be a list.

poolqc/pooling_qc_sample.py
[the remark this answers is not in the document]
callunity: Could this PR please be updated to be correct, and we can deal with the rebase issues later?

callunity:
Please rebase this branch on master and fix the conflicting files.

slazicoicr:
Remove `pandas.options.mode.chained_assignment = None`. Getting this warning is very useful and indicates possible problems with your code:

https://stackoverflow.com/questions/21463589/pandas-chained-assignments
https://www.dataquest.io/blog/settingwithcopywarning/

slazicoicr:
Use `dropna` rather than explicitly coverting `NaN` to str and iterating over it:

```
runs = df['Run'].sort_values(ascending=False).unique()
runs = [x for x in runs if str(x) != 'nan']
```

slazicoicr:
`run_URL_update` has the same bug as https://jira.oicr.on.ca/browse/GR-777?jql=text%20~%20dashi%20ORDER%20BY%20status%20DESC

Replace with the Query string fix and also allow the lane to be selected.

slazicoicr:
```
run = df[df['Run'] == run_alias]
run = run[~run['Run'].isna()]
```

I don't think the second is needed. It will only do anything if run_alias is NaN, which never occurs.

poolqc/pooling_qc_sample.py
```diff
@@ -144,98 +134,67 @@ def update_title(lane_value, run_value):
...
-)
-def Summary_table(run_alias, lane_alias):
-    run = df[(df['Run'] == run_alias) & (df['LaneNumber'] == lane_alias)]
-    run = run[~run['Run'].isna()].drop_duplicates('library')
-    run = run[~run['library'].isna()]
+def Summary_table(run):
+    # Adding 'on-the-fly' metrics
     run['% Mapped to Coding'] = run['Coding Bases'] / run['Passed Filter Aligned Bases'] * 100
     run['% Mapped to Intronic'] = run['Intronic Bases'] / run['Passed Filter Aligned Bases'] * 100
     run['% Mapped to Intergenic'] = run['Intergenic Bases'] / run['Passed Filter Aligned Bases'] * 100
```
slazicoicr: The `% Mapped` variables are already present from ETL, no need to calculate again.

poolqc/pooling_qc_sample.py
```diff
@@ -144,98 +134,67 @@ def update_title(lane_value, run_value):
...
-    return columns, data, style_data_conditional, csv, download
+    return columns, data, style_data_conditional, csv
 
 
-def update_sampleindices(run, rows, derived_virtual_selected_rows, colors):
+def update_sampleindices(run):
+    run = run.sort_values('library')
+    num_libraries = len(run['library'])
+    samples_passing_clusters = '%s/%s' % (sum(i > 20000 for i in run['SampleNumberReads']), num_libraries)
+    index_threshold = 50000000
```
slazicoicr: Put constants at the top of the file (outside a function), all-cap the variable name, and comment what they are.

ijafri26: This value will eventually be a variable, not a constant, since it will be something the client enters or will be calculated depending on the run itself. For now, this value is just a placeholder

callunity: Even though it's a placeholder for now, could you still move it to the top and comment it? This will hopefully make it easier to find later when we want to extract it or make it more dynamic.

PR #35, opened 2019-08-08
New Dashi layout ideas

slazicoicr:
* Using Material UI to improve look
* All filters are now in a hidden drawer that's shown with button click
* Graphs do not update until drawer is closed
* To increase rendering speed, option exists to select which graphs to display
* Graph shows loading animation while being rendered

### DIFF
--- a/dashi/rnaseqc/over_time.py
+++ b/dashi/rnaseqc/over_time.py
@@ -4,6 +4,7 @@ import dash_html_components as html
 import dash.dependencies as dep
 
 import plotly
+import sd_material_ui
 
 
 rna_df: pandas.DataFrame = pandas.read_hdf('./data/rnaseqqc_cache.hd5')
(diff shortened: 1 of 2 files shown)


dashi/rnaseqc/over_time.py
```diff
@@ -2,11 +2,13 @@
 import dash_core_components as dcc
 import dash_html_components as html
 import dash.dependencies as dep
+import dash.exceptions
+import itertools
 
 import plotly
+import sd_material_ui
```
callunity: I got a `ModuleNotFoundError: No module named 'sd_material_ui'` error, does this need to be added to `setup.py`?

ijafri26: yes, it can be installed with `pip install sd_material_ui`

slazicoicr: It does need to be added to `setup.py`. I will do so.

slazicoicr: @callunity I have added it to the `setup.py` file.

PR #39, opened 2019-08-21
Added extra input for update_url callback in Bcl2fastq

ijafri26:
Without this extra input, the callback doesn't fire, will investigate further on better ways to implement this.

Updated docstring for bcl2fastq

I have left in the print functions for confirmation of the callback firing 
Jira ticket: https://jira.oicr.on.ca/browse/GR-755


slazicoicr:
Implemented in #44

PR #40, opened 2019-09-05
RNASeQC improvements

slazicoicr:


CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.0] - 2019-09-05
```
callunity: I don't see the `0.2.0` tag anywhere, or a commit referencing it. Should these changes instead be labelled as `Unreleased`?

dashi/rnaseqc/over_time.py
```diff
@@ -51,72 +75,231 @@
...
+DEFAULT_TABLE_COLUMN = [
+    {'name': 'Library', 'id': 'Sample Name'},
+    {'name': 'Project', 'id': 'Study Title'},
+    {'name': 'Run', 'id': 'Sequencer Run Name'},
+    {'name': 'Lane', 'id': 'Lane Number'},
+    {'name': 'Kit', 'id': 'geo_prep_kit'},
+    {'name': 'Library Source Type', 'id': 'geo_library_source_template_type'},
+    {'name': 'Tissue Origin', 'id': 'geo_tissue_origin'},
+    {'name': 'Tissue Type', 'id': 'geo_tissue_type'},
+    {'name': 'Tissue State', 'id': 'geo_tissue_preparation'},
```
callunity: `Tissue State` is referred to as `Tissue Material` in MISO; is there a need to use a different term here?

dashi/rnaseqc/over_time.py
```diff
@@ -51,72 +75,231 @@
...
+]
+
+# Which columns will the data table always have
+DEFAULT_TABLE_COLUMN = [
+    {'name': 'Library', 'id': 'Sample Name'},
+    {'name': 'Project', 'id': 'Study Title'},
+    {'name': 'Run', 'id': 'Sequencer Run Name'},
+    {'name': 'Lane', 'id': 'Lane Number'},
+    {'name': 'Kit', 'id': 'geo_prep_kit'},
+    {'name': 'Library Source Type', 'id': 'geo_library_source_template_type'},
```
callunity: `Library Source Type` is referred to as `Library Design` in MISO; is there a need to use a different term here?

dashi/rnaseqc/over_time.py
```diff
@@ -51,72 +75,231 @@
...
+    in order through the values and assigning them the next available property.
+    Property lists cannot run out, as it loops back to the beginning.
+
+    Args:
+        values: The values of assign properties to. Need to be unique
+        prop: The properties to assign to
 
-def create_plot_dict(df, variable, colours=COLOURS, show_legend=False):
+    Returns: A dictionary with keys being the values (project, tissue, kit) and
+        the dictionary values being the properties (color, shape).
```
callunity: `A dictionary with keys being the values` -> `A dictionary with keys being the DataFrame values` ?
`and the dictionary values being the properties` -> `and the dictionary values being the plotting properties` ?

dashi/rnaseqc/over_time.py
```diff
@@ -132,44 +315,102 @@ def create_subplot(rna_df, graph_list):
...
-        sd_material_ui.RaisedButton(id='filter_button', label='Filters'),
+        html.Div([
+            html.Div(
+                sd_material_ui.RaisedButton(
+                    id='filter_button', label='Options'
+                ), style={'display': 'inline-block', }
+            ),
+            html.Div(
+                sd_material_ui.RaisedButton(
+                    id='update_from_table_button', label='Update Order',
```
callunity: `Update Order` is vague. Does this mean `re-graph`?

dashi/rnaseqc/over_time.py
```diff
@@ -132,44 +315,102 @@ def create_subplot(rna_df, graph_list):
...
+                ), style={'display': 'inline-block', 'margin-left': '15px'}
+            ),
+        ]),
```
callunity: These RaisedButtons overlap slightly with the chart top, so they end up looking like tabs (and it's not immediately obvious that they're clickable). Can the chart be moved to eliminate the overlap?

dashi/rnaseqc/over_time.py
```diff
@@ -132,44 +315,102 @@ def create_subplot(rna_df, graph_list):
...
+            [dash_table.DataTable(
+                id='data_table',
+                columns=DEFAULT_TABLE_COLUMN + [
+                    {'name': i, 'id': i} for i in METRICS_TO_GRAPH[:4]
+                ],
+                data=RNA_DF.to_dict('records'),
+                page_size=50,
+                sort_action='native',
+                sort_mode='multi',
+                export_format='csv'
```
callunity: Is it possible to name the downloaded sheet? `Data` isn't very informative.

CHANGELOG.md
slazicoicr: I will tag it in Github. It is used in the `setup.py` file.

https://github.com/slazicoicr/dashi/blob/f87dd16e6ff00f2d8db322143755c6467177b9f8/setup.py#L5

I still want to keep specific versions, as it makes it easy to install a specific pre-release version for development.

callunity: That makes sense. We often do a specific commit just for the release so that it can easily be referenced in the history; for this project, if you were to do so, the only code change in that commit would be the version number in setup.py.

dashi/rnaseqc/over_time.py
slazicoicr: Not at this point. I have made a ticket https://github.com/plotly/dash-table/issues/575

dashi/rnaseqc/over_time.py
slazicoicr: Changed in bce269f

dashi/rnaseqc/over_time.py
slazicoicr: Changed in bce269f

dashi/rnaseqc/over_time.py
slazicoicr: Clarified in bce269f. Let me know if it is clear.

dashi/rnaseqc/over_time.py
slazicoicr: Changed to `Update Graphs` bce269f. Note this button is used to update the graphs when the sorting of the DataTable is changed. You can change the sorting by clicking the small arrows in the header. Clicking the button makes the graphs re-plot given that order.

dashi/rnaseqc/over_time.py
slazicoicr: Added a small margin at the bottom in bce269f

PR #44, opened 2019-10-22
GR-777

slazicoicr:


dashi/bcl2fastq/index_summary.py
```diff
@@ -27,75 +30,70 @@
...
-try:
-    from app import app
-except ModuleNotFoundError:
-    import dash
+def generate_layout(qs) -> html.Div:
+    """
+    Within the main layout division there are 4 groups: 1. the section with the
+    100% bar graph 2.  Nested subdivision of: a) pie graph at 24% b) bar graph
+    at 76% 3. Sample Run hidden 4. Pruned  Unknown Hidden Note: span can be used
+    for inline setting
```
callunity: I'm not sure if black will allow it, but can this list be reformatted to have the numbers on new lines? Reading this had me doing some back-tracking to figure out what was going on. Point 4 is particularly confusing; I'm not sure if that's supposed to be "Pruned/Unknown (hidden)" with "Note: ..." on another line, or something else entirely?

dashi/bcl2fastq/index_summary.py
```diff
@@ -196,82 +194,80 @@ def create_pie_chart(run, pruned, total_clusters):
...
-    """
+            functions update_known_index_bar, update_unknown_index_bar,
+            update_pie_chart's data value, and update_pie_chart's fraction value
+        """
+
+        run = index[index[index_col.Run] == run_alias]
+        run = run[run[index_col.ReadNumber] == 1]
+        run = run[~run[index_col.SampleID].isna()]
+        run = run.drop_duplicates([index_col.SampleID, index_col.LaneNumber])
+        run[COL_LIBRARY] = run[index_col.SampleID].str.extract(
```
apmasell: Can you flag this with a TODO to fix it. We shouldn't be extracting libraries from file names or sample IDs.

dashi/bcl2fastq/index_summary.py
slazicoicr: Done in 2252416

dashi/bcl2fastq/index_summary.py
slazicoicr: Done in 2252416. I thought this was an easy fix with Pinery Sample Provenance, but the 10X indices make it non-trivial. Not sure if qc-etl should be doing this.

dashi/bcl2fastq/index_summary.py
```diff
@@ -27,75 +30,83 @@
...
-            ]
-        ),
-    ]
-)
 
-try:
-    from app import app
-except ModuleNotFoundError:
-    import dash
+def generate_layout(qs) -> html.Div:
```
avarsava: Contents of 'qs' variable was unclear to me. Had to dig into the code to figure out it's the search term grabbed from the URL. Documentation of source of 'qs' (and why the first character must be removed?) would be appreciated

PR #45, opened 2019-11-06
Migrate reports somewhat successfully

avarsava:
All but poolqc and bamqc_overtime ("shiny") work. Broken reports have been hidden from main menu but should still be accessible by URL.

Aside from rather adjustments to platonic architecture, report design and logic is unchanged (/plots directory is copied wholesale). Memoization is applied without finesse


slazicoicr:
bamqc_overtime ("shiny") works for me (takes a while to load and is known to glitch due to the large amount of data). I get a 404 error with http://127.0.0.1:5000/pooling_qc

PR #46, opened 2019-11-12
Refactor pd-migration

callunity:
I've reorganized the pages into a `views` folder so they don't get lost among the routes and utils.

(removing pre-commit linting from this PR)

### DIFF
--- a/application/dash_application/dash_routes.py
+++ b/application/dash_application/dash_routes.py
@@ -1,20 +1,20 @@
 from dash import Dash
-from dash.dependencies import Input, Output
-import dash_html_components as html
 
-## Adds Dash to provided Flask server
+
+# Adds Dash to provided Flask server
 def add_dash(server):
-    ## Create Dash instance using Flask object as server
-    dash_app = Dash(__name__, server=server) 
+    # Create Dash instance using Flask object as server
+    dash_app = Dash(__name__, server=server)
 
-    ## Load callbacks into memory
+    # Set initial Dash page's layout
+    from . import known_pages_router
+    known_pages_router.init_callbacks(dash_app)
+    dash_app.layout = known_pages_router.layout
+
+    # Load all callbacks into memory
     from . import pages
     for page in pages.pages:
         page.init_callbacks(dash_app)
 
-    ## Set your initial Dash page's layout to this variable 
-    from . views import known_pages_router
-    dash_app.layout = known_pages_router.layout
-
-    ## Return the server object from Dash to overwrite Flask server object
+    # Return the server object from Dash to overwrite Flask server object
     return dash_app.server
--- a/application/dash_application/dash_routes.py
+++ b/application/dash_application/dash_routes.py
@@ -13,8 +13,8 @@ def add_dash(server):
         page.init_callbacks(dash_app)
 
     ## Set your initial Dash page's layout to this variable 
-    from . import url_handler
-    dash_app.layout = url_handler.layout
+    from .views import index_page
+    dash_app.layout = index_page.layout
 
     ## Return the server object from Dash to overwrite Flask server object
     return dash_app.server
--- a/application/dash_application/dash_routes.py
+++ b/application/dash_application/dash_routes.py
@@ -13,8 +13,8 @@ def add_dash(server):
         page.init_callbacks(dash_app)
 
     ## Set your initial Dash page's layout to this variable 
-    from .views import index_page
-    dash_app.layout = index_page.layout
+    from . views import known_pages_router
+    dash_app.layout = known_pages_router.layout
 
     ## Return the server object from Dash to overwrite Flask server object
     return dash_app.server
--- a/application/dash_application/views/bamqc_gbovertime.py
+++ b/application/dash_application/views/bamqc_gbovertime.py
@@ -36,7 +36,10 @@ layout = html.Div(
 
 def init_callbacks(dash_app):
     @dash_app.callback(
-        Output(ids["month_plot"], "figure"), [Input(ids["lib"], "value")]
+        Output(ids["month_plot"], "figure"),
+        [
+            Input(ids["lib"], "value")
+        ]
     )
     @dash_app.server.cache.memoize(timeout=60)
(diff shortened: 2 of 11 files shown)


application/dash_application/views/bamqc_gbovertime.py
```diff
@@ -26,24 +21,23 @@
...
+        core.Graph(id=ids["cum_plot"]),
     ]
 )
 
+
 def init_callbacks(dash_app):
     @dash_app.callback(
-        Output(ids['month_plot'], "figure"), 
-        [Input(ids['lib'], "value")])
+        Output(ids["month_plot"], "figure"), [Input(ids["lib"], "value")]
```
avarsava: Is there any way to tweak the code formatter's behaviour? I think having the Input array to the same line as Output reduces the readability of callbacks, and it seems inconsistent about when it does this, too

application/dash_application/views/bcl2fastq.py
```diff
@@ -170,17 +169,16 @@ def update_layout(run_alias):
...
-            run, pruned, total_clusters
-        )
+        pie_data, textarea_fraction = create_pie_chart(run, pruned, total_clusters)
 
         return (
             create_known_index_bar(run),
             create_unknown_index_bar(pruned),
             pie_data,
-            textarea_fraction
+            textarea_fraction,
```
avarsava: extra comma?

application/dash_application/views/runreport.py
```diff
@@ -12,12 +12,7 @@
 
 page_name = "runreport/proj_hist"
 
-ids = init_ids([
-    'project',
-    'focused_run',
-    'coverage_dist',
-    'click-data'
-])
+ids = init_ids(["project", "focused_run", "coverage_dist", "click-data"])
```
avarsava: Here too, I find having all the items on one line harder to read

application/dash_application/views/bcl2fastq.py
callunity: Yes, to minimize the diff. If more items are added to the structure, it will be a one-line change to add the new item as opposed to two lines, one with added comma and one with the new item.

application/dash_application/views/bamqc_gbovertime.py
callunity: Black's philosophical underpinning is that things like this simply aren't configurable. I tried to jiujitsu the formatter by adding `#` + newline between input/output items but it still reformatted them onto the same line (with one `#` at the end of the line, sigh). I do agree that it makes the inputs and outputs more difficult to read, which is concerning given how central they are to dashi. I'd like to look into other automatic formatters that don't sacrifice readability. @slazicoicr did you investigate any other formatters when you were looking at Black?

slazicoicr: Not too deeply, as Black worked well for gsiqcetl out of the box. There is:
* autopep8 
* yapf

There is a roundup at https://medium.com/3yourmind/auto-formatters-for-python-8925065f9505 which does state that yapf is way more configurable than black.

callunity: Ok. I'm going to take a look at the formatters and see if we can't find something that maintains readability.

slazicoicr:
I get the following exception when I try to start the application:

```
python wsgi.py

Traceback (most recent call last):
  File "wsgi.py", line 4, in <module>
    app = create_app()
  File "/home/slazic/Documents/Code/2018-08-27-Dash_trial/application/__init__.py", line 26, in create_app
    app = dash_routes.add_dash(app)
  File "/home/slazic/Documents/Code/2018-08-27-Dash_trial/application/dash_application/dash_routes.py", line 11, in add_dash
    from . import pages
  File "/home/slazic/Documents/Code/2018-08-27-Dash_trial/application/dash_application/pages.py", line 23, in <module>
    pages.append(importlib.import_module(prefix + name))
  File "/home/slazic/.local/miniconda3/envs/dash/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ModuleNotFoundError: No module named 'application.dash_application.views.url_handler'
```

callunity:
@slazicoicr good point. Fixed now.

application/dash_application/dash_routes.py
```diff
@@ -13,8 +13,8 @@ def add_dash(server):
         page.init_callbacks(dash_app)
 
     ## Set your initial Dash page's layout to this variable 
-    from . import url_handler
-    dash_app.layout = url_handler.layout
+    from .views import index_page
```
avarsava: It seems like changing . to .views is the whole of the fix here. I'm not sure I agree with changing 'url_handler' to 'index_page' because that makes it sound like it plays the role of 'application/templates/index.html'. url_handler/index_page doesn't actually have any content of its own it displays

callunity: Hrm, looking at it more closely it does look like a router for known pages, as opposed to a view. Having the layout element tricked me into thinking that it would display something. It sounds like it could make sense to rename `url_handler/index_page` to `known_pages_router` and to clearly comment what its function is (and keep it out of the `views` folder). Sound ok @avarsava ?

avarsava: I thought `url_handler` described what it did (take action based on incoming URL values), but i'm also OK with `known_pages_router` if you think that makes more sense :) and more documentation is always good

callunity: Does it handle all URL-related changes, like query strings or anchors?

avarsava: it _could_ , by virtue of all Dash callbacks existing in memory at once, but that's not really the intended pattern

callunity: Then I will update it to `known_pages_router` since it's not intended to handle all URL-related changes, since that's an important disambiguation in my brain and is acceptable to yours :)

application/dash_application/views/runreport.py
callunity: updated

PR #49, opened 2019-11-27
Improve UI appearance

apmasell:


application/dash_application/views/pre_exome.py
```diff
@@ -191,7 +191,9 @@ def generateDebugLine(click, runs, firstsort, secondsort, colourby,
 layout = html.Div(className='body',
     children=[
         navbar("Pre-Exome"),
-        html.Div(className='sidebar',
+        html.Div(className='row flex-container',
```
avarsava: Could there be an explanation (or link to explanation) of these class names somewhere?

avarsava:
This is going to sound insane, but in rshiny that 'terminal output' textarea expands to the height of its contents. Could we do that here?

avarsava:
Also could the sidebar be as long as it needs to be to show all the controls rather than scrolling?

application/dash_application/views/pre_exome.py
apmasell: Honestly, I copied it from the template. I'm not entirely sure what these really do.

apmasell:
This should address everything except the loading problem, which seems to be a client-side issue where React is clearing the loading message and then spending forever rendering stuff.

PR #50, opened 2019-11-29
Wrap in docker container

morgantaschuk:


README.md
```diff
@@ -1,6 +1,43 @@
...
+
+
+1. Ensure your ssh key has been added to OICR's Bitbucket and you can access and
+   clone
+   [gsi-qc-etl](https://bitbucket.oicr.on.ca/projects/GSI/repos/gsi-qc-etl/browse).
+2. Download the gsi-qc-etl cache data to `cache_files` (or modify
+   docker-compose.yml to point to the correct location). The current location for
+   this on OICR's cluster is at
+   `/.mounts/labs/mcphersonlab/public/slazic/2019-01-16-GSI_ETL/dev_cache_files`.
+3. Build the container with `docker-compose build`. Note that this completes
```
callunity: I tried to build this locally and ran into an error:
```
  git@bitbucket.oicr.on.ca: Permission denied (publickey).
  fatal: Could not read from remote repository.

  Please make sure you have the correct access rights and the repository exists.
```
I tried adding my key to the "Access Keys" tab on Bitbucket and was told that this key was already associated with user harmstrong...

morgantaschuk: You don't want to add it to Access keys. That's for machines. Can you clone that repository with the ssh url?

callunity: Yes, I'm able to clone with the ssh URL.

morgantaschuk: We figured out that the issue arose because @callunity had a passphrase on her SSH key. Will need to figure out how to pass that in.

README.md
```diff
@@ -1,6 +1,43 @@
...
+   installation of gsi-qc-etl before launching the app.
+4. Launch with `docker-compose up`.`
```
avarsava: I completed all the steps up to this point, but this fails with` ModuleNotFoundError: No module named 'gsiqcetl'` from pre_exome.py

callunity: Did everything earlier in the installation succeed? When I ran into issues with my SSH key, this is the step that it failed at but earlier in the logs it showed that the gsi-qc-etl repository did not successfully download, which led to this eventual failure.

avarsava: `docker-compose build` says it built successfully

morgantaschuk: @avarsava it's `docker-compose run` that installs the gsi-qc-etl dependency

avarsava: since the last commit this step now fails with 
```ERROR: for dashi  Cannot create container for service dashi: invalid mount config for type "bind": bind source path does not exist: /home/avarsava/eclipse-workspace/dashi/.mongopass```
is there a file missing from this PR?

Comment (morgantaschuk) [reply README.md]: Oh yes, sorry. I pushed in a hurry last night. there should be a file `.mongopass` in your `dashi` directory with the Mongo DB password. I whispered it at you in Slack.

Comment (morgantaschuk) [reply README.md]: You'll have to create it.

Comment (avarsava) [reply README.md]: Thanks! Should the README be updated to include this instruction?

Comment (morgantaschuk) [reply README.md]: Yes, will do so today.

File: .docker/requirements_less.txt [hunk 3]
```diff
@@ -0,0 +1,11 @@
+Flask_Caching>=1.7.2
```
callunity: I'm leery of adding a second `requirements.txt` file as I'm not sure we'll remember to update it. What do you think about generating on the fly with something like awk to remove the gsi-qc-etl line?

.docker/start.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# mount your private key for bitbucket into the container to allow this command to succeed
+pip install --trusted-host pypi.python.org -r requirements.txt
+
+export PINERY_URL='http://seqbio-pinery-prod-www.hpc.oicr.on.ca:8080/pinery-ws-miso/'
+export GSI_ETL_CACHE_DIR='/cache'
+export MONGO_URL="mongodb://provenance_ro:$(cat /run/secrets/mongo_password)@provenance-mongo-db.gsi.oicr.on.ca:27017/provenance"
+export GSI_QC_ETL_RUNSCANNER_MASTER_URL='http://runscanner.res.oicr.on.ca:8080/'
```
callunity: It turns out the run scanner URL should be `http://runscanner.res.oicr.on.ca:8080/runs/all`. (You could omit it here if you wanted since it's only used for building the cache; Dash should only use prebuilt caches) (Savo is updating the gsi-qc-etl documentation to reflect this)

README.md
callunity: It works for me now that I've removed the passphrase from my SSH key.

README.md
```diff
@@ -1,6 +1,43 @@
...
+* Access to [gsi-qc-etl](https://bitbucket.oicr.on.ca/projects/GSI/repos/gsi-qc-etl/browse)
+
+
+1. Ensure your ssh key has been added to OICR's Bitbucket and you can access and
+   clone
+   [gsi-qc-etl](https://bitbucket.oicr.on.ca/projects/GSI/repos/gsi-qc-etl/browse).
+2. Download the gsi-qc-etl cache data to `cache_files` (or modify
+   docker-compose.yml to point to the correct location). The current location for
+   this on OICR's cluster is at
+   `/.mounts/labs/mcphersonlab/public/slazic/2019-01-16-GSI_ETL/dev_cache_files`.
```
callunity: This address is accurate for the v0.5.0 cache. Caches that match current `gsi-qc-etl@master` schemas are available at `/scratch2/groups/gsi/<development or production>/qcetl`

docker-compose.yml
```diff
@@ -28,8 +28,12 @@ services:
...
+#      - ssh_passphrase
 
-
+# If your SSH key requires a passphrase, uncomment out the `ssh_passphrase` lines in the dashi service and in the secrets section below
+# Create a file at .passphrase that contains your passphrase. No other changes are necessary.
 secrets:
   mongo_password:
     file: .mongopass
+#  ssh_passphrase:
+#    file: .passphrase
```
callunity: Can `.passphrase` be added to `.gitignore`?

README.md
morgantaschuk: There's instructions in the README and compose on how to use an SSH key with a passphrase.

README.md
morgantaschuk: I've added instructions on working with a passphrase.

.docker/start.sh
morgantaschuk: Removed.

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    build: 
+      context: .
+      dockerfile: Dockerfile
+    restart: always
+    ports:
+      - '5000:5000'
+    volumes:
+      # location of gsi-qc-etl cache files
+      - type: bind
+        source: ./cache_files
```
callunity: It would be nice if this could be not hardcoded; the location of this is stored in the `$GSI_QC_ETL_ROOT_DIRECTORY` variable present on the host machine. This is not a need-to-have though. Everything else works as-is.

README.md
```diff
@@ -1,6 +1,71 @@
...
+    [https://bitbucket.oicr.on.ca/projects/GSI/repos/gsi-qc-etl/browse](https://bitbucket.oicr.on.ca/projects/GSI/repos/gsi-qc-etl/browse)
+2. Check that your SSH key is located in `~/.ssh/`. If it is not, change the
+    bind paths in `docker-compose.yml` for id_rsa, id_rsa.pub, and known_hosts.
+3. Check if your SSH key requires a passphrase. If it is prompting from inside
+    the container, it will fail. Create a private file with your SSH key
+    passphrase. Open `docker-compose.yml` and uncomment out the
+    `ssh_passphrase` lines in the dashi service and in the `secrets` section,
+    and point docker-compose to your passphrase file in the secrets section.
+    When next launching the container, the start.sh script will automatically
+    provide the passphrase upon request.
```
avarsava: After uncommenting these lines after an initial failure, docker-compose build and up still fail until i launch them interactively, I think because that forces a clean build (build keeps saying it's using the cache). I think rerunning with `--no-cache` may be necessary

morgantaschuk: I'll add a note.

docker-compose.yml
morgantaschuk: Changed it to point to `$HOME/qcetl`.

PR #51, opened 2019-12-02
Add bones of pre-RNA graph

callunity:
Has working graphs, but lacks the DataTable. I'm working on that now.


application/dash_application/plot_builder.py
```diff
@@ -1,68 +1,108 @@
...
+#                 'asterisk', 'asterisk-open', 'hash',
+#                 'hash-open', 'hash-dot', 'hash-open-dot',
+#                 'y-up', 'y-up-open', 'y-down',
+#                 'y-down-open', 'y-left', 'y-left-open',
+#                 'y-right', 'y-right-open', 'line-ew',
+#                 'line-ew-open', 'line-ns', 'line-ns-open',
+#                 'line-ne', 'line-ne-open', 'line-nw',
+#                 'line-nw-open']
+
+ALL_SYMBOLS = [
```
avarsava: this list got commented out then added back in?

application/dash_application/views/preqc_rna.py
```diff
@@ -0,0 +1,531 @@
...
+            State(ids['shape-by'], 'value'),
+        ]
+    )
+    def update_pressed(click,
+                       runs,
+                       first_sort,
+                       second_sort,
+                       colour_by,
+                       shape_by):
+        print("App initiated")  # TODO: delete me
```
apmasell: Delete me?

callunity: Deleted now :disappointed_relieved:

application/dash_application/views/preqc_rna.py
```diff
@@ -0,0 +1,531 @@
...
+        rna_df[RNA_COL.TotalReads] / pow(10, 6), 3)
+    rna_df[special_cols["Percent Correct Strand Reads"]] = round(
+        rna_df[RNA_COL.CorrectStrandReads] * 100
+    )
+
+    # Pull in sample metadata from Pinery. Keep only the columns we care about
+    pinery_samples = util.get_pinery_samples_from_active_projects()
+    pinery_samples = pinery_samples[
+        [
+            "DV200", # FIXME
```
avarsava: FIXME? here and at 119

application/dash_application/views/preqc_rna.py
```diff
@@ -0,0 +1,531 @@
...
+}
+
+
+def fill_in_shape_col(df: DataFrame, shape_col: str):
+    all_shapes = get_shapes_for_values(shape_or_colour_values[
+                                           shape_col].tolist())
+    # for each row,
+    df[special_cols['shape']] = df.apply(lambda row: all_shapes.get(row[
+        shape_col]), axis=1)
+    return df
```
avarsava: would it be possible for shape functionality move to plot_builder or df_manipulation? that way scattergl() and scatter_graph() might not have to replicate the generate() in plot_builder as well

application/dash_application/views/preqc_rna.py
callunity: Yes. By all rights it looks like it should be `PINERY_COL.DV200` but that resolves to `"dv200"` and I get a "no column with this name" error. The error doesn't throw when using the literal `"DV200"`. I haven't yet figured it out and it's not mission-critical at this time so I'm keeping the FIXME and it's high on my to-do list

application/dash_application/plot_builder.py
callunity: Yes -- sorry, tried to make that clearer in the commit message for this PR but should have added it to the file instead. Will update.

application/dash_application/views/preqc_rna.py
callunity: Yes, that makes sense.

callunity: (Back story: I developed the scattergl() and scatter_graph() functions while you were working on the plot_builder, and haven't had the time to reimplement them using the plot_builder yet. It's also high on my to-do list)

PR #53, opened 2019-12-02
Add basics of pre-WGS

apmasell:
This needs:

* better input data
* sliders that work
* to use Alexis's graph generation code

### DIFF
--- a/application/dash_application/views/preqc_wgs.py
+++ b/application/dash_application/views/preqc_wgs.py
@@ -1,0 +1,477 @@
+import dash_core_components as core
+import dash_html_components as html
+import numpy
+import plotly.graph_objects as go
+from dash.dependencies import Input, Output, State
+
+import gsiqcetl.column
+import pinery
+from gsiqcetl import QCETLCache
+from . import navbar
+from ..dash_id import init_ids
+from ..plot_builder import get_shapes_for_values, fill_in_shape_col
+from ..utility import df_manipulation as util
+
+""" Set up elements needed for page """
+page_name = "preqc-wgs"
+
+ids = init_ids([
+    # Buttons
+    "update-button",
+    "download-button",
+
+    # Sidebar controls
+    "run-id-list",
+    "first-sort",
+    "second-sort",
+    "colour-by",
+    "shape-by",
+    "passed-filter-reads-slider",
+
+    # Graphs
+    "total-reads",
+    "mean-insert",
+    "duplication",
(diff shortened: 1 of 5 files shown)


callunity:
(I'm currently updating the RNA code to use Alexis's graph generation code, and am happy to help convert this one as well)

apmasell:
I'll take a stab at it. I was mostly concerned about getting the right BamQC data.

application/templates/index.html
```diff
@@ -6,7 +6,9 @@
 	<body>
 		<h1>Dashi</h1>
 		<h2>PRE-EXOME</h2>
-		<h3><a href="/preexome">PRE-EXOME Report</a></h3>
+		<h3><a href="/preqc-exome">PRE-EXOME Report</a></h3>
```
avarsava: did the URL for the pre-exome report change?

application/dash_application/views/preqc_wgs.py
```diff
@@ -0,0 +1,475 @@
...
+    cache = QCETLCache()
+    if True:  # DELETE(amasella): once IchorCNA data is available
+        wgs_df = cache.bamqc.bamqc
+        wgs_df[ICHOR_COL.Ploidy] = numpy.NaN
+        wgs_df[ICHOR_COL.TumorFraction] = numpy.NaN
+        # This doesn't seem to exist even though it should
+        wgs_df[BAMQC_COL.MarkDuplicates_PERCENT_DUPLICATION] = numpy.NaN
+    else:
+        wgs_df = cache.bamqc.bamqc.merge(cache.ichorcna.ichorcna[[ICHOR_COL.Ploidy, ICHOR_COL.TumorFraction]], how="left", left_on=[
+            BAMQC_COL.Run, BAMQC_COL.Lane, BAMQC_COL.Barcodes], right_on=[ICHOR_COL.Run, ICHOR_COL.Lane, ICHOR_COL.Barcodes])
```
callunity: Could this be spaced out a bit more (one item per line) for readability?

application/templates/index.html
callunity: ^ Good point. @apmasell if you're making this change, the `page_name` will have to be changed in `pre_exome.py`

apmasell: I'm still very confused associations between URL, file name, `page_name` and that list.

callunity: We should probably standardize them

PR #54, opened 2019-12-02
pre-RNA now with data table

callunity:


application/dash_application/plot_builder.py
```diff
@@ -84,7 +85,8 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, hovertext
             )
         )
     traces = []
-    grouped_data = sorted_data.groupby(colourby) #TODO: is this inefficient?
+    grouped_data = sorted_data.groupby([colourby, shapeby]) #TODO: is this
```
avarsava: Could you please explain why you group by shapeby as well?

application/dash_application/utility/df_manipulation.py
```diff
@@ -63,13 +65,23 @@ def get_pinery_samples_from_active_projects():
 
 def df_with_pinery_samples(df: DataFrame, pinery_samples: DataFrame, ius_cols:
                            List[str]):
-    """Do an outer merge between the DataFrame and Pinery samples data."""
-    return df.merge(
+    """Do an outer merge between the DataFrame and modern Pinery samples
```
apmasell: Comment is out of date.

application/dash_application/plot_builder.py
callunity: (We spoke offline and determined that `groupby` will soon be made obsolete)

PR #55, opened 2019-12-03
Upgrade etl preexome

avarsava:
no other changes

### DIFF
--- a/application/dash_application/views/pre_exome.py
+++ b/application/dash_application/views/pre_exome.py
@@ -345,22 +345,22 @@ layout = core.Loading(fullscreen=True, type="cube", children=[html.Div(className
         html.Div(className='seven columns',
             children=[
                 core.Graph(id=ids['total-reads'],
-                    figure=generateTotalReads(bamqc, bamqc[BAMQC_COL.Sample].str[0:4], 'none')
+                    figure=generateTotalReads(bamqc, bamqc[PINERY_COL.StudyTitle], 'none')
                 ),
                 core.Graph(id=ids['unmapped-reads'],
-                    figure=generateUnmappedReads(bamqc, bamqc[BAMQC_COL.Sample].str[0:4], 'none')
+                    figure=generateUnmappedReads(bamqc, bamqc[PINERY_COL.StudyTitle], 'none')
                 ),
                 core.Graph(id=ids['non-primary-reads'],
-                    figure=generateNonprimaryReads(bamqc, bamqc[BAMQC_COL.Sample].str[0:4], 'none')
+                    figure=generateNonprimaryReads(bamqc, bamqc[PINERY_COL.StudyTitle], 'none')
                 ),
                 core.Graph(id=ids['on-target-reads'],
-                    figure=generateOnTargetReads(bamqc, bamqc[BAMQC_COL.Sample].str[0:4], 'none')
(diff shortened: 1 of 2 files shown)


application/dash_application/views/pre_exome.py
```diff
@@ -5,15 +5,14 @@
...
-
-# TODO filter down to MiSeq runs only?
-bamqc_cols = gsiqcetl.load.bamqc_columns('v1')
-bamqc = gsiqcetl.load.bamqc('v1').sort_values(by=[bamqc_cols.Sample, bamqc_cols.TotalReads], ascending=False)
-
+from gsiqcetl import QCETLCache
+from gsiqcetl.column import BamQcColumn
+import pinery
 
+# # TODO filter down to MiSeq runs only?
```
callunity: At the last meeting we decided we wouldn't filter down to MiSeq runs by default, but we may later add a toggle for only displaying MiSeq runs/filtering by instrument model.

application/dash_application/views/pre_exome.py
```diff
@@ -48,6 +47,28 @@
...
+    bamqc_df = util.df_with_normalized_ius_columns(bamqc_df, BAMQC_COL.Run, BAMQC_COL.Lane, BAMQC_COL.Barcodes)
+
+    pinery_samples = util.get_pinery_samples_from_active_projects()
+    # TODO filter??
+
+    bamqc_df = util.df_with_pinery_samples(bamqc_df, pinery_samples, util.bamqc_ius_columns)
+
+    bamqc_df = util.df_with_instrument_model(bamqc_df, PINERY_COL.SequencerRunName)
+
+    return bamqc_df
```
callunity: I think this will include all WG samples as well. Could you add a `util.filter_by_library_design` call to select just the `EX` and `TS` samples?

application/dash_application/views/pre_exome.py
```diff
@@ -392,36 +413,36 @@ def updatePressed(click,
             passedfilter):
 
         # Apply get selected runs
-        data = bamqc[bamqc[bamqc_cols.Run].isin(runs)]
-        data[bamqc_cols.GroupID] = data[bamqc_cols.GroupID].fillna("")
+        data = bamqc[bamqc[BAMQC_COL.Run].isin(runs)]
+        data[BAMQC_COL.GroupID] = data[BAMQC_COL.GroupID].fillna("")
```
callunity: The `get_pinery_samples` function from earlier takes care of the `fillna` for GroupID

application/dash_application/views/pre_exome.py
```diff
@@ -392,36 +413,36 @@ def updatePressed(click,
...
         elif secondsort == 'BAMQC_READSPERSTARTPOINT':
-            sortby.append(bamqc_cols.ReadsPerStartPoint)
+            sortby.append(BAMQC_COL.ReadsPerStartPoint)
 
         if colourby == 'run':
-            colourby_strategy = bamqc_cols.Run
+            colourby_strategy = BAMQC_COL.Run
         elif colourby == 'project':
-            colourby_strategy = data[bamqc_cols.Sample].str[0:4]
+            colourby_strategy = data[BAMQC_COL.Sample].str[0:4]
```
callunity: Please use `data[PINERY_COL.StudyTitle]`

avarsava:
@callunity all addressed

PR #58, opened 2019-12-03
Extra updates

callunity:
* Standardizes keys
  * Fewer expensive calls to modify Pinery samples
  * Fill in shape column

### DIFF
--- a/application/dash_application/utility/df_manipulation.py
+++ b/application/dash_application/utility/df_manipulation.py
@@ -53,14 +53,22 @@ _active_samples = _pinery_samples.loc[_pinery_samples[
     PINERY_COL.StudyTitle].isin(
     _active_projects)]
 
+_runs_with_instruments = _runs.copy(deep=True).merge(
+        _instruments[[INSTRUMENTS_COL.ModelName, INSTRUMENTS_COL.Platform,
+                      INSTRUMENTS_COL.InstrumentID]],
+        how="left",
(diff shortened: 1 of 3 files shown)


application/dash_application/utility/df_manipulation.py
```diff
@@ -26,41 +26,41 @@
...
 
 def get_pinery_samples_from_active_projects():
-    pinery_samples = get_pinery_samples()
-    active_projects = _projects.loc[_projects[PROJECT_COL.IsActive] == True]
-    active_projects = active_projects[PROJECT_COL.Name].unique()
-    active_samples = pinery_samples.loc[pinery_samples[
-        PINERY_COL.StudyTitle].isin(
-        active_projects)]
-    return active_samples
+    return _active_samples
```
avarsava: At this point, could you remove the `_` from the beginning of the variable name and let developers get the variable directly?

callunity: Hm.. Given how we'll be using the samples across graphing modules, I had wanted to not return the original samples dataframes. It looks like I didn't implement that properly though. What do you think about instead changing `get_active_samples` and `get_pinery_samples` so they return copies of the dataframes instead of the original ones, so the original ones can't be modified between graphing modules?

avarsava: That sounds good.

callunity: Updated

PR #59, opened 2019-12-04
Small updates

callunity:
RNA "Second sort" values are copied from second sort values in TGL Shiny.

### DIFF
--- a/application/dash_application/views/__init__.py
+++ b/application/dash_application/views/__init__.py
@@ -6,11 +6,12 @@ def navbar(current):
         children=[
             dbc.DropdownMenu(
                 children=[
-                    dbc.DropdownMenuItem("Exome", href="exome", style={"fontSize": "12pt"}),
-                    dbc.DropdownMenuItem("Pre-Exome", href="preexome", style={"fontSize": "12pt"}),
-                    dbc.DropdownMenuItem("sWGS", href="swgs", style={"fontSize": "12pt"}),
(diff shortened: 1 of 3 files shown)


application/dash_application/views/__init__.py
```diff
@@ -6,11 +6,12 @@ def navbar(current):
         children=[
             dbc.DropdownMenu(
                 children=[
-                    dbc.DropdownMenuItem("Exome", href="exome", style={"fontSize": "12pt"}),
-                    dbc.DropdownMenuItem("Pre-Exome", href="preexome", style={"fontSize": "12pt"}),
-                    dbc.DropdownMenuItem("sWGS", href="swgs", style={"fontSize": "12pt"}),
-                    dbc.DropdownMenuItem("Pre-WGS", href="prewgs", style={"fontSize": "12pt"}),
-                    dbc.DropdownMenuItem("CFMEDIP", href="cfmedip", style={"fontSize": "12pt"}),
+                    dbc.DropdownMenuItem("Pre-Exome", href="preqc-exome",
```
avarsava: for future consideration - I think we could generate this list by iterating over the page_names from pages.py so we don't have to update this manually

apmasell: It would be nice to do `index.html` the same way.

avarsava: `index.html` is I think meant to be rewritten with some nice looking icons eventually...

PR #62, opened 2019-12-04
Update pre-WGS graphing code and all Second Sort options

callunity:
GR-918 & GR-917


application/dash_application/views/preqc_wgs.py
```diff
@@ -158,99 +158,108 @@ def get_wgs_data():
...
+    )
 
 
 def generate_on_target_reads(df, colour_by, shape_by):
-    return scatter_graph(
-        df, colour_by, shape_by, PINERY_COL.SampleName,
-        special_cols["On-target Reads"],
-        "On-target Reads", "Percent (%)")
+    return generate(
+        "On-target Reads Reads",
```
avarsava: Minus one 'Reads'?

callunity: yes! Fixed.

PR #63, opened 2019-12-04
[GR-911] Start with blank Run ID box

avarsava:


avarsava:
Updated to include blank graphs as well, also a bugfix that was causing all runs to show at all times in PRE-EXOME report

application/dash_application/views/preqc_wgs.py
```diff
@@ -157,6 +157,8 @@ def get_wgs_data():
 WGS_DF = fill_in_shape_col(WGS_DF, initial_shape_col, shape_or_colour_values)
 WGS_DF = fill_in_colour_col(WGS_DF, initial_colour_col, shape_or_colour_values)
 
+EMPTY_WGS = pd.DataFrame(columns=WGS_DF.columns)
```
callunity: Error: `pd` is not defined

avarsava: fixed

avarsava:
> Is there any way to keep the run list open after a run is added, until the user clicks away? This would help if they end up having to scroll down the list to get to a set of runs of interest.

Doesn't look like it's available out of the box (https://dash.plot.ly/dash-core-components/dropdown). May be able to hack something, but we'll have to decide on priority
 
> When I click "Update" _only on the pre-exome page_, I get the following warning in the logs:
> 
> ```
> /home/harmstrong/workspace/dashi/application/dash_application/plot_builder.py:72: SettingWithCopyWarning:
> 
> 
> A value is trying to be set on a copy of a slice from a DataFrame.
> Try using .loc[row_indexer,col_indexer] = value instead
> 
> See the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
> ```

Working on this

callunity:
> > Is there any way to keep the run list open after a run is added, until the user clicks away? This would help if they end up having to scroll down the list to get to a set of runs of interest.
> 
> Doesn't look like it's available out of the box (https://dash.plot.ly/dash-core-components/dropdown). May be able to hack something, but we'll have to decide on priority

That makes sense. I'll make a ticket and we can decide on priority later.

avarsava:
> > > Is there any way to keep the run list open after a run is added, until the user clicks away? This would help if they end up having to scroll down the list to get to a set of runs of interest.
> > 
> > 
> > Doesn't look like it's available out of the box (https://dash.plot.ly/dash-core-components/dropdown). May be able to hack something, but we'll have to decide on priority
> 
> That makes sense. I'll make a ticket and we can decide on priority later.

:+1: 

Still working on clearing that slicing error, will merge once that's done

PR #64, opened 2019-12-04
Specify table columns

callunity:
GR-914

### DIFF
--- a/application/dash_application/table_builder.py
+++ b/application/dash_application/table_builder.py
@@ -1,0 +1,27 @@
+from typing import List
+
+import dash_table as tabl
+from pandas import DataFrame
+
+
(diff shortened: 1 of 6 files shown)


application/dash_application/table_builder.py
```diff
@@ -0,0 +1,27 @@
...
+    # we have no QC data for these records
+    df = df[~df[filter_on].isna()]
+    return tabl.DataTable(
+        id=table_id,
+        columns=[{"name": col, "id": col} for col in columns],
+        data=df.to_dict('records'),
+        export_format="csv",
+        style_data_conditional=[
+            {
+                "if": {"row_index": "odd"},
```
apmasell: Yikes. This is awful as a convention.

callunity: I know :cry:

PR #65, opened 2019-12-05
[GR-911] 'Add All' button

avarsava:
Adds all the runs to the Run ID box. Doesn't update the graphs, but still triggers the loading screen (I suppose any callback will).

### DIFF
--- a/application/dash_application/views/pre_exome.py
+++ b/application/dash_application/views/pre_exome.py
@@ -263,7 +263,6 @@ layout = core.Loading(fullscreen=True, type="cube", children=[html.Div(className
                 html.Br(),
 
                 html.Button('Add All', id=ids["all"]),
-                html.Button('Clear All', id=ids["clear"]),
                 html.Br(),
 
                 html.Label([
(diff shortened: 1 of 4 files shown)


callunity:
Could the button be moved to be inline with or below the "Run" label? That way, it'll be clear that it applies only to this table (which will be more relevant when we've got several multiselect dropdowns ie. for runs, projects, instrument models).

Could the styling also be changed so it doesn't look like the Update and Download buttons which have  [global/more-dramatic] effects?

application/dash_application/views/pre_exome.py
```diff
@@ -18,6 +18,7 @@
     # Buttons
     'update-button',
     'download-button',
+    'all',
```
callunity: Perhaps `all-runs` instead of just `all`?

avarsava: done

avarsava:
> Could the button be moved to be inline with or below the "Run" label? That way, it'll be clear that it applies only to this dropdown (which will be more relevant when we've got several multiselect dropdowns ie. for runs, projects, instrument models).
> 
> Could the styling also be changed so it doesn't look like the Update and Download buttons which have [global/more-dramatic] effects?

I've added a bit of CSS, please let me know what you think.

PR #66, opened 2019-12-05
Move frange and percentageOf to slider_utils

avarsava:
This finally bothered me enough to fix


application/dash_application/utility/slider_utils.py
```diff
@@ -0,0 +1,14 @@
...
+def frange(min, max, step):
+    range = []
+    i = min
+    while i <= max:
+        range.append(round(i, 2))
+        i += step
+    return range
+
+def percentageOf(data, bamqc_column):
+    return (data[bamqc_column] / data[BAMQC_COL.TotalReads]) * 100
```
callunity: What do you think about changing `bamqc_column` to `numerator_column` and `BAMQC_COL.TotalReads` to `denominator_col` (and passing `denominator_col` as another arg to the function)? That would allow you to use it for data sources other than BamQC. (I'm not attached to `numerator_col` and `denominator_col` as names)

avarsava: Done!

PR #69, opened 2019-12-05
Fix overzealous click event

avarsava:
Putting Buttons inside of Labels causes the whole Label to watch for clicks on behalf of the Button.

### DIFF
--- a/application/dash_application/views/pre_exome.py
+++ b/application/dash_application/views/pre_exome.py
@@ -282,10 +282,10 @@ layout = core.Loading(fullscreen=True, type="cube", children=[html.Div(className
                     )
                 ]), html.Br(),
 
-                html.Label([
-                    "Search Sample:",
-                    core.Input(id=ids['search-sample'])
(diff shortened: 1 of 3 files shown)


application/dash_application/views/pre_exome.py
```diff
@@ -283,10 +282,10 @@ def generate_mean_insert_size(current_data, colourby, shapeby, shownames,
...
                 ]), html.Br(),
 
-                # html.Label([
-                #     "Search Sample:",
-                #     core.Input(id=ids['search-sample'])
-                # ]), html.Br(),
+                html.Label([
+                    "Search Sample:",
+                    core.Input(id=ids['search-sample'])
+                ]), html.Br(),
```
callunity: Looks like these got uncommented a bit early, but should be fine to merge as I know you're working on this next

avarsava: Oops! Fixed

PR #72, opened 2019-12-06
Filter by instruments and projects

callunity:
GR-910 & GR-912


avarsava:
Pressing 'All Runs' and 'All Instruments' and then 'Update' on PRE-EXOME produces graphs, but selecting 'All Projects' then 'Update' clears the graphs?

application/dash_application/views/pre_exome.py
```diff
@@ -412,10 +452,17 @@ def update_pressed(click,
...
         else:
-            data = bamqc[bamqc[BAMQC_COL.Run].isin(runs)]
+            data = bamqc
+
+        if runs:
+            data = data[data[BAMQC_COL.Run].isin(runs)]
+        if instruments:
+            data = data[data[INSTRUMENT_COLS.ModelName].isin(instruments)]
+        if projects:
+            data = data[data[PINERY_COL.StudyTitle].isin(instruments)]
```
avarsava: I think this should be `.isin(projects)`

callunity: I think you're right! Will update

callunity: updated

avarsava:
I'm not sure I understand the filtering by looking at the front end. If I have 3 runs selected, whether i have all projects or no projects selected doesn't seem to affect the graphs? if I don't understand it our users might not either

callunity:
My reasoning was that users will likely want to answer a specific question like "Show me all MiSeq" data. Rather than have them click "All Runs", scroll down, select "MiSeq", scroll down, click "All Projects", scroll up, click "Update", I thought it would be easier to assume that if they're entering one filtering criteria, they likely want to filter on only that data and display everything else. 

If we graph only what they select, if they select "MiSeq" and click "Update" with no runs selected and no projects selected graphs nothing, which doesn't seem very useful. I agree that graphing something that isn't filled out is  not necessarily the most intuitive though, and I'll think further on how we could display this.

apmasell:
As another weird case, I'm wondering if the filtering sets need to be updated. If I select all the MiSeq runs, should I still have the option to enter a project that's never been on a MiSeq? Admittedly, this is complicated, so I'm happy with it as is.

callunity:
> As another weird case, I'm wondering if the filtering sets need to be updated. If I select all the MiSeq runs, should I still have the option to enter a project that's never been on a MiSeq? Admittedly, this is complicated, so I'm happy with it as is.

That's an interesting question! My first thought is that it's more obvious what's happening when the flow is 1) select project, select MiSeq, click graph -> 0 samples displayed, rather than 2) select project, select MiSeq, "wait, I thought I selected a project, where'd it go". I'll add this to the list of UI things to ponder.

PR #73, opened 2019-12-09
[GR-909] Sample Search

avarsava:
Select one or more samples to make very large and bright red.

### DIFF
--- a/application/dash_application/plot_builder.py
+++ b/application/dash_application/plot_builder.py
@@ -63,6 +63,7 @@ PLOTLY_DEFAULT_COLOURS=[
     '#17becf'   # blue-teal
 ]
 
+BIG_MARKER_SIZE = 20
 
 def fill_in_shape_col(df: DataFrame, shape_col: str, shape_or_colour_values:
         dict):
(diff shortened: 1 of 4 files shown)


application/dash_application/plot_builder.py
```diff
@@ -75,19 +75,28 @@ def fill_in_shape_col(df: DataFrame, shape_col: str, shape_or_colour_values:
...
                                             colour_col].tolist())
         # for each row,
         df['colour'] = df.apply(lambda row: all_colours.get(row[
             colour_col]), axis=1)
+
+    if highlight_samples:
+        df.loc[df[PINERY_COL.SampleName].isin(highlight_samples), 'colour'] = '#F00'
+    return df
+
+def fill_in_size_col(df: DataFrame):
```
callunity: I'm a little leery of having this depend on the colour, as it'll introduce unexpected behaviour if this function happens to be called before `fill_in_colour_col`. Could this also make use of the `highlight_samples` parameter?

application/dash_application/plot_builder.py
```diff
@@ -149,6 +159,18 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
             line={"width": 1, "color": "black", "dash": "dash"},
             name="Cutoff"
         ))
+    if not highlight_df.empty:
+        traces.append(go.Scattergl( # Draw highlighted items on top
+            x=x_fn(highlight_df),
+            y=y_fn(highlight_df),
+            name="Highlighted Samples",
```
callunity: Can the name of the sample be displayed instead?

application/dash_application/plot_builder.py
avarsava: Fixed

application/dash_application/plot_builder.py
avarsava: Currently all the markers are drawn in one trace, which only allows for 1 legend entry. I'll look into looping over them and whether that's too slow for high # of highlighted samples

callunity:
The highlighted samples are pretty huge, and when a large number of samples are displayed, they block out significant portions of the graph around them. Could they be made somewhat smaller?

avarsava:
> The highlighted samples are pretty huge, and when a large number of samples are displayed, they block out significant portions of the graph around them. Could they be made somewhat smaller?

HI @callunity , I've made them less huge, please let me know what you think

application/dash_application/plot_builder.py
avarsava: So it seems as though when looping over samples one at a time (rather than in groups of them) it doesn't understand what to do when given a Sample name for the x co-ordinate. I think this feature might need further discussion.

avarsava:
> The "Highlighted Samples" don't seem to be showing at all on pre-WGS or pre-RNA, either on the graphs or in the legend.

Latest fix should address this

avarsava:
After discussion with @callunity , spun off expanding the legend to ticket GR-955.

PR #74, opened 2019-12-09
Create tables for samples below cutoffs

apmasell:


apmasell:
It's intentionally not showing all the samples, but only the ones with failures; if everything passes, the sample is excluded.

There's no pre-built collapsible sections, but there are tabs. Does it make sense to flip between the failures and all the data (and that gives them headers too)?

Did you want the failing values in those cells too?

callunity:
Ah, then in that case I like the idea of tabs, as I can't imagine them wanting to view both tables at once, and it gives us convenient labelling of the tables.

I agree, users will probably want to see failing values in the cells

apmasell:
Okay. Tabs and values added.

callunity:
The "Raw Data" tab takes 3+ seconds to load when all runs are selected, and there's no indication to the user that something is happening during this time. Does it make things clearer to the user if the tables are wrapped in Loading spinners?

Also, the tables are displaying as monospaced on pre-WGS and regular font on pre-RNA and pre-EX. Can this be standardized? I'm partial to regular font.

apmasell:
Fixed the fonts, added spinners, and I moved more of the functionality into reused code.

callunity:
Fonts look great and I like the code reuse, but I'm not seeing the spinners -- it just looks like nothing is happening for 3-4 seconds after clicking "Raw Data" (on pre-WGS + all runs).

apmasell:
Apparently you're not supposed to use tabs without callbacks. You get no loading animations and slow switching between tabs. https://github.com/plotly/dash-core-components/issues/256

PR #77, opened 2019-12-11
Deal with NaN in failed samples table

apmasell:


application/dash_application/table_builder.py
```diff
@@ -77,9 +80,14 @@ def cutoff_table(table_id: str, data: DataFrame, limits: List[Tuple[str, str, fl
             },
             *({
                 "if": {"column_id": name, "filter_query": "{%s} contains 'Failed'" % name},
-                "backgroundColor": "pink"
+                "backgroundColor": "mistyrose"
+
+            } for (name, _, _) in limits),
+            *({
+                "if": {"column_id": name, "filter_query": "{%s} = 'Missing'" % name},
+                "backgroundColor": "mintcream"
```
callunity: Can this colour be changed? It's the same saturation and intensity as the grey used for alternating rows (see attached) for nearly all forms of colourblindness.
redblind

apmasell: Is now papayawhip

PR #81, opened 2019-12-12
RPSP fails if above cutoff; fix slider legend

callunity:
GR-951


application/dash_application/table_builder.py
```diff
@@ -82,12 +86,12 @@ def cutoff_table(table_id: str, data: DataFrame, limits: List[Tuple[str, str, fl
                 "if": {"column_id": name, "filter_query": "{%s} contains 'Failed'" % name},
                 "backgroundColor": "mistyrose"
 
-            } for (name, _, _) in limits),
+            } for (name, _, _, _) in limits),
```
slazicoicr: Can change to `for (name, *_) in limits)` to avoid dealing with element counting

callunity:
> I still get strange behaviour with RPSP. For 191029_A00469_0066_AHJWGHDMXX, both runs have RPSP below 1, but when I set the cutoff to 1, both fail.

Is this for the pre-exome report? I'm seeing the reads per start point as being 1.644 and 1.337 for those samples, so they should fail when the cutoff is at 1. It looks like the values in the fail table are being rounded down to the nearest integer though, and I will change that to show decimals.

PR #82, opened 2019-12-12
Fixed up plot titles and y-axis

slazicoicr:
GR-954


application/dash_application/views/preqc_rna.py
```diff
@@ -212,7 +212,7 @@ def generate_reads_per_start_point(df, colour_by, shape_by, cutoff):
         df,
         lambda d: d[PINERY_COL.SampleName],
         lambda d: d[RNA_COL.ReadsPerStartPoint],
-        "",
+        "Fraction",
```
callunity: Can the RPSP Y axis not be labelled, rather than call it "Fraction"? As you said, it's :woman_shrugging: as to why they call it that

application/dash_application/views/preqc_rna.py
```diff
@@ -195,11 +195,11 @@ def generate_total_reads(df, colour_by, shape_by, cutoff):
 
 def generate_unique_reads(df, colour_by, shape_by):
     return generate(
-        "Unique Reads (Passed Filter)",
+        "Unique Passed Filter Reads (%)",
```
callunity: In graphs below you use `% X`, whereas here and in a couple of other graphs you use `X (%)`. Can this be standardized?

application/dash_application/views/preqc_rna.py
slazicoicr: Done

application/dash_application/views/preqc_rna.py
slazicoicr: Done

PR #84, opened 2019-12-13
[GR-916] Make shape_or_colour_values consistent

avarsava:


avarsava:
@callunity rebased on master

application/dash_application/views/preqc_wgs.py
```diff
@@ -100,6 +100,7 @@
     {"label": "Run", "value": PINERY_COL.SequencerRunName},
     {"label": "Kit", "value": PINERY_COL.PrepKit},
     {"label": "Tissue Prep", "value": PINERY_COL.TissuePreparation},
+    {"label": "Library Design", "value": PINERY_COL.LibrarySourceTemplateType},
```
callunity: Could Library Design please be removed for this report? By definition, it'll only display WG samples.

avarsava: Done!

PR #93, opened 2019-12-19
v0.1 pr

callunity:
I'm not tied to "-SNAPSHOT" but figure we can keep it in here for now and [revisit later](https://jira.oicr.on.ca/browse/GR-979)


apmasell:
I've not seen `-SNAPSHOT` outside of Java. :man_shrugging:

callunity:
Yes, it's definitely a deliberate Java-ism and I'm certain that more appropriate tools/conventions exist within the Python ecosystem. There is the linked ticket to investigate and make a better decision for the future.

PR #97, opened 2020-01-07
[GR-967] Log all the params when update is called

avarsava:
Redirects all logging to dashi.log, including the 'Running on 127.0.0.1' message, so watch dashi.log for that

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ venv
 .idea
 .mongopass
 .passphrase
+dashi.log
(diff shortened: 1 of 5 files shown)


apmasell (approved):

callunity (commented):
Could these be logged as JSON? This would make parsing easier even without ELK.

Please add `dashi.log` to `.gitignore`

avarsava:
> Could these be logged as JSON? This would make parsing easier even without ELK.
> 
> Please add `dashi.log` to `.gitignore`

Both done!

callunity (approved):

PR #101, opened 2020-01-10
Request data of runs from last X days

callunity:
GR-947
- allow URL to determine which filters (run dates) get applied to page on load
- refactor pages to use template


application/dash_application/known_pages_router.py
```diff
@@ -1,34 +1,92 @@
...
+                                    href=link,
+                                    external_link=True,
+                                    disabled=current == label,
+                                    style={"fontSize": "12pt"})
+    return dbc.NavbarSimple(
+        id="navbar",
+        children=[
+            dbc.DropdownMenu(
+                children=[
+                    menu_item("Pre-Exome", "preqc-exome"),
```
apmasell: You could build this from `pages_info` if inclined.

apmasell (approved):

application/dash_application/known_pages_router.py
callunity: That's much more elegant.

avarsava (approved):

PR #105, opened 2020-01-14
Allow specifying use of QC-ETL@master

callunity:
GR-994

### DIFF
--- a/.docker/start.sh
+++ b/.docker/start.sh
@@ -7,11 +7,10 @@ then
     /dashi/.docker/enter_passphrase $(cat /run/secrets/ssh_passphrase)
 fi
 
-echo "USE_BLEEDING_EDGE_ETL: $USE_BLEEDING_EDGE_ETL"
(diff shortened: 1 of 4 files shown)


apmasell (approved):

avarsava (approved):

callunity:
(This didn't work as I had hoped, so the PR will remain open while I troubleshoot)

callunity:
Updated. To test, add `USE_BLEEDING_EDGE_ETL=1` in your `.env` file, then do `docker-compose build && docker-compose up`. In the output from docker-compose up you should see that it installs `gsi-qc-etl@master` instead of the `gsi-qc-etl@v0.16.1` present in `requirements.txt`.

avarsava:
Should this environment variable be documented in the README?

callunity:
> Should this environment variable be documented in the README?

It could be. Right now the README only documents the required environment variables; I'll update it to describe all the possible environment variables (I'll include the one from #110 so @slazicoicr and I aren't writing over top of each other)

PR #109, opened 2020-01-15
[GR-1003] None bugfix in known_pages_router

slazicoicr:


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and as of version 1.0.0, follows semantic versioning.
 
 ## [Unreleased]
+### Changed
+  * `nav_handler` and `content_handler` no longer throw exception on `None`
```
callunity: I worry that this will be pretty opaque to us in a few months. What about a message that explains the "why", something like `Do not throw an exception on empty path`? Also, can this go under a `### Fixed` header instead of `### Changed`?

avarsava (approved):

callunity (approved):

PR #111, opened 2020-01-15
[GR-967] end_date adjustments for log aggregation

avarsava:
Convert end_date to date (vs datetime) and only log it if there has been a change to its value. This will improve our ability to aggregate logs to view trends in use of filters.

### DIFF
--- a/application/dash_application/known_pages_router.py
+++ b/application/dash_application/known_pages_router.py
@@ -52,7 +52,7 @@ layout = html.Div([
     core.Location(id='url', refresh=False),
     navbar(default_title),
     core.Loading(id='page-content', type='dot'),
-    html.Footer(id='footer', children=[html.Hr(), "Dash version {0}".format(version)])
+    html.Footer(id='footer', children=[html.Hr(), "Dash version {0} | Data version ".format(version), html.Span(id='data-version')])
 ])
 
 
--- a/application/dash_application/known_pages_router.py
+++ b/application/dash_application/known_pages_router.py
@@ -51,8 +51,8 @@ def navbar(current):
 layout = html.Div([
     core.Location(id='url', refresh=False),
     navbar(default_title),
-    core.Loading(id='page-content'),
-    html.Footer(id='footer', children=[html.Hr(), "Dash version {0} | Data version ".format(version), html.Span(id='data-version')])
+    core.Loading(id='page-content', type='dot'),
+    html.Footer(id='footer', children=[html.Hr(), "Dash version {0}".format(version)])
 ])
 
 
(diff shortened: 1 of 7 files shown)


callunity (approved):
If you wanted, it looks like you could factor the logging function into a separate file and import that file/function into the views pages (so future similar changes would only be on one file).

apmasell (approved):

avarsava:
Refactored for abstraction, integrated Heather's all_x feature, and fixed missing callback and NaN runs on TS and RNA views

application/dash_application/known_pages_router.py
```diff
@@ -51,8 +51,8 @@ def menu_item(label, link):
 layout = html.Div([
     core.Location(id='url', refresh=False),
     navbar(default_title),
-    core.Loading(id='page-content'),
-    html.Footer(id='footer', children=[html.Hr(), "Dash version {0} | Data version ".format(version), html.Span(id='data-version')])
+    core.Loading(id='page-content', type='dot'),
+    html.Footer(id='footer', children=[html.Hr(), "Dash version {0}".format(version)])
```
callunity: Maybe a bad rebase -- could the footer code from develop please be retained?

application/dash_application/utility/log_utils.py
```diff
@@ -0,0 +1,30 @@
...
+import datetime
+import logging
+import json
+import pdb
+
+compare = lambda s, t: Counter(s) == Counter(t)
+
+def collapse_if_all_selected(selected_items: List[str], all_items: List[str], all_title: str) -> List[str]:
+    if compare(selected_items, all_items):
+        return [all_title] # Why array?
```
callunity: I put it in as an array because `selected_items` is an array, and figured it might be easier to manipulate these in logging if they're all of the same type. You're more experienced in the logging manipulation than I am though -- what do you think?

application/dash_application/known_pages_router.py
avarsava: Should be good now

application/dash_application/views/single_lane_wgs.py
```diff
@@ -591,3 +601,10 @@ def all_projects_requested(click):
     )
     def all_kits_requested(click):
         return [x for x in ALL_KITS]
+
+    @dash_app.callback(
+        Output(ids['library-designs-list'], 'value'),
+        [Input(ids['all-library-designs'], 'n_clicks')]
+    )
+    def all_library_designs_requested(click):
```
callunity, line 609: (I spent too much time failing to figure out why this didn't work. Glad you succeeded! :laughing: )

callunity (approved):

application/dash_application/utility/log_utils.py
avarsava: Updated the comment to reflect the decision to return an array for the sake of format consistency

apmasell (approved):

PR #112, opened 2020-01-15
Allow specifying a log file location

callunity:
Removes app namespacing of variables, as isolation should be handled by the virtualenv and .env file.

### DIFF
--- a/wsgi.py
+++ b/wsgi.py
@@ -5,13 +5,13 @@ import dotenv
 
 dotenv.load_dotenv()
 
-logging.basicConfig(filename='dashi.log', level=logging.INFO)
+logging.basicConfig(filename=(os.getenv("LOG_FILE_LOCATION") or "dashi.log"), level=logging.INFO)
 
-if os.getenv("DASHI_LOG_TO_CONSOLE") == "True":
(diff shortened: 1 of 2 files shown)


wsgi.py
```diff
@@ -5,13 +5,13 @@
 
 dotenv.load_dotenv()
 
-logging.basicConfig(filename='dashi.log', level=logging.INFO)
+logging.basicConfig(filename=(os.getenv("LOG_FILE_LOCATION") or "dashi.log"), level=logging.INFO)
```
slazicoicr, line 8: README has the default location at `~/logs/dashi.log`, but this would save it where the Dashi app was started.

callunity: The README indicates an example location, not a default location. I can update the README to indicate the default location as well, and change this to match that default location.

slazicoicr (approved):

wsgi.py
callunity: Updated the README to indicate that the file is created in the directory where the app was started (as I didn't want to make any assumptions about file systems and `~/`...)

avarsava (approved):

PR #113, opened 2020-01-17
Show data versions

apmasell:


avarsava (approved):

callunity (commented):
This will add a second footer to the page, as one is already created in `known_pages_router.py`. It's not the most elegant way to display it, but I'm not fundamentally opposed. However, I think some descriptive text would help give context, something like "Date(s) of data generation: $DATAVERSION"?

apmasell:
Yeah, I didn't see an easy way to plug into the existing footer.

callunity (approved):

PR #114, opened 2020-01-23
[GR-1011] Only plot samples that have QC data

slazicoicr:
Currently, all samples in Pinery are plotted, even when they don't have any QC data. This has several problems:

- Filter drop down menus include Runs, Projects, Kits that have no data to plot (confusing to user)
- Slows down Dashi, as it has to handle many rows, but most of them hold no data
- Makes graphs mostly white space
Screenshot from 2020-01-23 11-12-11

Fix only keeps Pinery entries that have QC associated with them:
Screenshot from 2020-01-23 11-14-00

The DataTable now excludes samples with no QC data. I think that makes sense, as Dashi deals with QC, not with samples having a Pinery entry.

### DIFF
--- a/application/dash_application/utility/df_manipulation.py
+++ b/application/dash_application/utility/df_manipulation.py
@@ -123,7 +123,7 @@ def get_pinery_samples_from_active_projects():
 def df_with_pinery_samples(df: DataFrame, pinery_samples: DataFrame, ius_cols:
                            List[str]):
     """Do an left merge between the DataFrame and modern Pinery samples
-    data. Only samples in DataFrame will be kept."""
+    data. Only samples in QC DataFrame will be kept."""
     df = df.merge(
         pinery_samples,
         how="left",
--- a/application/dash_application/utility/df_manipulation.py
+++ b/application/dash_application/utility/df_manipulation.py
@@ -122,11 +122,11 @@ def get_pinery_samples_from_active_projects():
 
 def df_with_pinery_samples(df: DataFrame, pinery_samples: DataFrame, ius_cols:
                            List[str]):
-    """Do an outer merge between the DataFrame and modern Pinery samples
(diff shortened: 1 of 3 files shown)


application/dash_application/utility/df_manipulation.py
```diff
@@ -122,11 +122,11 @@ def get_pinery_samples_from_active_projects():
 
 def df_with_pinery_samples(df: DataFrame, pinery_samples: DataFrame, ius_cols:
                            List[str]):
-    """Do an outer merge between the DataFrame and modern Pinery samples
-    data."""
+    """Do an left merge between the DataFrame and modern Pinery samples
+    data. Only samples in DataFrame will be kept."""
```
callunity: Can `Only samples in DataFrame will be kept.` be changed to `Only samples in QC DataFrame will be kept.`?

callunity (changes requested):
This will break the single_lane_wgs page, as it merges bamqc3 + Pinery, then ichorcna + (bamqc3 + Pinery). Can you please update this page to do an outer merge on bamqc3 + ichorcna, then (bamqc3 + ichorcna) + Pinery?

callunity (approved):

application/dash_application/utility/df_manipulation.py
```diff
@@ -122,11 +122,11 @@ def get_pinery_samples_from_active_projects():
 
 def df_with_pinery_samples(df: DataFrame, pinery_samples: DataFrame, ius_cols:
                            List[str]):
-    """Do an outer merge between the DataFrame and modern Pinery samples
-    data."""
+    """Do an left merge between the DataFrame and modern Pinery samples
```
avarsava: 'an left' -> 'a left'

avarsava (approved):

PR #115, opened 2020-01-23
[GR-1012] Fixed exception due to Dash callback output mismatch

slazicoicr:


avarsava (approved):

application/dash_application/known_pages_router.py
```diff
@@ -83,9 +83,9 @@ def nav_handler(path):
         ])
     def content_handler(path, qs):
         if path == '/None' or path is None:
-            return '404'
+            return None, '404'
```
callunity: Can the order of these please be flipped? The `page-content` should remain the string `404` (for `404 Not Found`, which might be the more useful message to display) and `None`  should be the `data-version`.

callunity (approved):

PR #116, opened 2020-01-23
[GR-1014] Show Names can display multiple fields in hover label

slazicoicr:


application/dash_application/plot_builder.py
```diff
@@ -134,7 +134,7 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
     if colourby == shapeby:
         name_format = lambda n: "{0}".format(n[0])
     else:
-        name_format = lambda n: "{0} {1}".format(n[0], n[1])
+        name_format = lambda n: "{0}<br>{1}".format(n[0], n[1])
```
avarsava, line 184: \<br\> -> \<br /\>

avarsava (approved):

callunity (commented):
Can the fields have a default order that they are displayed in? Right now if I select everything, the tissue origin and tissue type are separated by the tissue preparation line, which is unintuitive. I'd propose the following order:
```
Sample
Tissue Origin
Tissue Type
Group ID
Tissue Prep
Kit
Run
```

Since Group ID can be blank and Tissue Prep and Kit can both be "Unspecified", I think it could make sense to label those lines, ie.
```
...
Br
P
Group ID: 
Tissue Preparation: Unspecified
Kit: Unspecified
200123_A00469_0123_BCKDHELYXX
```

Since the hover text is no longer just displaying names, perhaps the dropdown label could be changed from "Show Names" to "Data Labels" or similar? It might also be worth having a "Select All Labels" button.

avarsava:
(accidentally only looked at latest commit)

application/dash_application/plot_builder.py
slazicoicr: The `<br>` is a Plotly thing: https://help.plot.ly/adding-HTML-and-links-to-charts/#step-2-the-essentials

application/dash_application/views/single_lane_rna.py
```diff
@@ -602,3 +605,12 @@ def all_kits_requested(click):
...
     def all_library_designs_requested(click):
         return [x for x in ALL_LIBRARY_DESIGNS]
+
+    @dash_app.callback(
+        Output(ids['show-data-labels'], 'value'),
+        [Input(ids['show-all-data-labels'], 'n_clicks')],
+        [State(ids['show-data-labels'], 'options')]
+    )
+    def all_data_labels_requested(click, avail_options):
+        if click is not None:
```
avarsava, line 615: Why the check for `None` on click?

slazicoicr: This callback fires when the page is loaded (`click` is None), meaning that the drop box gets filled in with all values by default. I do not understand why this does not occur with the other "Fill in All` buttons.

callunity: ^ I also noticed this behaviour and am working on a consistent fix (that involves `if click is not None` everywhere)

callunity (approved):

PR #117, opened 2020-01-28
GR-1006 query by run

callunity:
Allows the user to request a specific run by appending `?run=RUN_NAME` to the URL. 
Also involves a significant refactoring to be able to use a single function to manipulate the dataframe on initial page load and also when updating the page.


application/dash_application/utility/log_utils.py
```diff
@@ -24,6 +24,6 @@ def collapse_all_params(params, collapsing_functions):
 def log_filters(params, collapsing_functions, logger):
     collapse_all_params(params, collapsing_functions)
     del params['click']
-    if datetime.datetime.strptime(params['end_date'], '%Y-%m-%d').date() == datetime.date.today():
+    if datetime.datetime.strptime(params['end_date'].split("T")[0], '%Y-%m-%d').date() == datetime.date.today():
```
avarsava: why the return to datetimes for end_date?

avarsava (commented):
What is the proper syntax to use both 'run=' and 'last='? I've tried separating them with ',' and '&' but neither seems to work

callunity:
> What is the proper syntax to use both 'run=' and 'last='? I've tried separating them with ',' and '&' but neither seems to work

Hmm, I'm able to get it to work by separating them with `&`. Which values are you using for `run` & `last`?

avarsava:
> > What is the proper syntax to use both 'run=' and 'last='? I've tried separating them with ',' and '&' but neither seems to work
> 
> Hmm, I'm able to get it to work by separating them with `&`. Which values are you using for `run` & `last`?

I guess chalk this one up to user error, because it works just fine for me on a fourth try

application/dash_application/utility/log_utils.py
callunity: :sweat_smile: long story short, I went down a rabbit hole of dates and datetimes and this was the only thing I could get to work yesterday. But your comment prompted me to look at it with fresh eyes today, and I see that the run range input is empty by default and not filled in like I had thought it was, so I have reverted that (1a8e4c2)!

avarsava (approved):

apmasell (approved):

PR #120, opened 2020-02-04
[GR-830] Make Dash graph page for EXOME files

asofmosq:


application/dash_application/views/call_ready_ts.py
```diff
@@ -0,0 +1,579 @@
...
+
+
+# todo: come back and include whatever else should be a special column
+special_cols = {
+    "Total Reads (Passed Filter)": "Total reads passed filter",
+    "Percent Unique Reads (PF)": "Percent unique reads",
+    "Callability (14x/8x)": "callability",
+    "Purity": "Purity",
+    "File SWID ichorCNA": "File SWID ichorCNA",
+    "File SWID MutectCallability": "File SWID MutectCallability"
```
callunity: Please add `File SWID BamQC3` and `File SWID HsMetrics`

application/dash_application/views/call_ready_ts.py
```diff
@@ -0,0 +1,579 @@
...
+}
+
+ts_table_columns = [*BAMQC_COL.values(), *HSMETRICS_COL.values(), *ICHOR_COL.values(), *CALL_COL.values(),
+                    *special_cols.values()]
+
+
+def get_merged_ts_data():
+    """"
+    Come back and fill in docstring
```
callunity: Please fill in

application/dash_application/views/call_ready_ts.py
```diff
@@ -0,0 +1,579 @@
...
+
+    bamqc3_df[special_cols["Total Reads (Passed Filter)"]] = round(
+        bamqc3_df[BAMQC_COL.TotalReads] / 1e6, 3)
+    ichorcna_df[special_cols["Purity"]] = round(
+        ichorcna_df[ICHOR_COL.TumorFraction] * 100.0, 3)
+    callability_df[special_cols["Callability (14x/8x)"]] = round(
+        callability_df[CALL_COL.Callability] * 100.0, 3)
+
+    ichorcna_df.rename(columns={ICHOR_COL.FileSWID: special_cols["File SWID ichorCNA"]}, inplace=True)
+    callability_df.rename(columns={CALL_COL.FileSWID: special_cols["File SWID MutectCallability"]}, inplace=True)
```
callunity, line 126: Please also rename bamqc_df FileSWID and hsmetrics_df FileSWID

application/dash_application/views/call_ready_ts.py
```diff
@@ -0,0 +1,579 @@
...
+
+    # Join IchorCNA+HsMetrics and Callability data
+    ts_df = ts_df.merge(
+        callability_df,
+        how="outer",
+        left_on=util.ichorcna_merged_columns,
+        right_on=util.callability_merged_columns)
+
+    # Join QC data and Pinery data
+    ts_df = util.df_with_pinery_samples_merged(ts_df, pinery_samples, util.ichorcna_merged_columns)
```
callunity: This join needs to happen after the BamQC3 merge (because otherwise it'll drop all Pinery samples which have no hsmetrics/ichorcna/callability data, but those samples might have bamqc3 data -- this might be why you're not seeing any BamQC metrics)

callunity:
I'm running into the "Too many WebGL contexts used, must cannibalize some and therefore won't display all your graphs" issue in Chrome. This happens when we display more than eight graphs, so I'll add this to next week's agenda to see if we can consolidate some graphs or display some non-interactively.

callunity (approved):

application/dash_application/views/call_ready_ts.py
```diff
@@ -0,0 +1,591 @@
...
+                                                         # "value": },
+                                                         {"label": "Mean Target Coverage",
+                                                          "value": HSMETRICS_COL.MeanTargetCoverage},
+                                                         {"label": "Callability",
+                                                          "value": CALL_COL.Callability},
+                                                         {"label": "Tumor Fraction",
+                                                          "value": ICHOR_COL.TumorFraction},
+                                                         {"label": "HS Library Size",
+                                                          "value": HSMETRICS_COL.HsLibrarySize},
+                                                         {"label": "PctExcDupe??",
```
avarsava: "??"

avarsava (approved):
I can't seem to launch it ("KeyError: 'Unknown cache name: bamqc3merged'") but :+1: if this is just my env missing something

application/dash_application/views/call_ready_ts.py
callunity: Ah, should be "Percent Exact Duplicates"

callunity:
@avarsava what version of gsi-qc-etl do you have installed in your venv? It should be 0.18.0 or higher; if it's at master then you'll want to do a `pip install -r requirements.txt --upgrade`

avarsava:
> @avarsava what version of gsi-qc-etl do you have installed in your venv? It should be 0.18.0 or higher; if it's at master then you'll want to do a `pip install -r requirements.txt --upgrade`

Interesting, my requirements.txt says 0.17.0 but I'm up to date on `develop`

callunity:
Hmm. I could be wrong, then. My open PR upgrades it to v0.19.0 anyway though, so maybe try upgrading yours locally, safe in the knowledge that it will be upgraded soon?

PR #121, opened 2020-02-04
More updates to Call-Ready WGS page

callunity:
GR-848
* highlight samples works
* fail tables display data correctly (and can distinguish between tumour & normal cutoffs)
* includes the BamQC3 cache, but in order to test it, you'll need to do the following (until the caches are rebuilt with gsiqcetl@v0.19.0, in which case it'll be business as usual):
  * `pip install -r requirements.txt --upgrade`
  * `vim [v]env/lib/python3.6/site-packages/gsiqcetl/bamqc3/__init__.py` and comment out lines 97-99 and 103 (the ones referencing `TotalBasesOnTarget`)

Once Angie's call-ready TS PR is merged, I'll rebase this onto it and apply the cutoff refactoring to the call-ready TS page.


application/dash_application/views/call_ready_wgs.py
```diff
@@ -172,43 +192,33 @@ def get_merged_wgs_data():
...
         PINERY_COL.RootSampleName)
 
 
 def generate_mean_coverage(df, graph_params):
-    # TODO: should display two cutoffs: cutoff_coverage_tumour, cutoff_coverage_normal 
-    return "https://jira.oicr.on.ca/browse/GR-848?focusedCommentId=160788&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-160788"
+    return generate(
+        "Mean Coverage", df,
+        lambda d: d[PINERY_COL.RootSampleName],
+        lambda d: d[special_cols["Mean Coverage"],
```
asofmosq: Just missing a close bracket here

asofmosq (approved):

slazicoicr (commented):
* There are Coverage cutoffs, but no Coverage graph
* Filter by Library Designs only has the WG option. If this will always be the only option, that filter box can be removed
* Unique Reads is always 100%. ETL has all `non primary reads` at 0. My guess is that the merged pipeline discards secondary reads. This makes this plot pointless and it can be removed.

callunity:
> * There are Coverage cutoffs, but no Coverage graph
> * Filter by Library Designs only has the WG option. If this will always be the only option, that filter box can be removed
> * Unique Reads is always 100%. ETL has all `non primary reads` at 0. My guess is that the merged pipeline discards secondary reads. This makes this plot pointless and it can be removed.

* The coverage graph is waiting on ETL rebuilding a bamqc3merged cache with the necessary column (TotalBasesOnTarget); that should be happening within the next hour.
* Filter by Library Designs will have whatever data the input caches have. At this point it's only WG data, but NN, AS, CH, and CM data will also be displayed on this page and once we have data for these libraries, the library designs will be added to the filter list.
* Rather than discard the requested graph entirely, I'd prefer if we discussed this at the next reporting meeting to see if it's something that should be fixed in our pipeline.

slazicoicr (approved):

callunity:
I've updated the PR to apply my changes to the cutoff tables to the call-ready TS page as well; hidden the unique reads table for now; and displayed the Coverage table.

slazicoicr:
In Call Ready WGS `Unmapped Reads (%)`, the Y-axis is read counts.
In Call Ready TS, I think we are hitting the maximum graph bug, as the top two graphs are empty for me.

callunity:
> In Call Ready WGS `Unmapped Reads (%)`, the Y-axis is read counts.

[Updated](https://github.com/oicr-gsi/dashi/pull/121/commits/56d9c726a86d06565dc08b27261b2917710ccdec)

> In Call Ready TS, I think we are hitting the maximum graph bug, as the top two graphs are empty for me.

I agree, and we'll discuss options for dealing with this at the next reporting meeting.

PR #122, opened 2020-02-05
[GR-1039] Add Data Table to bcl2fastq view

slazicoicr:


avarsava (approved):

application/dash_application/views/bcl2fastq.py
```diff
@@ -89,6 +108,27 @@ def layout(qs):
...
             ),
+            html.Br(),
+            core.Tabs(id=ids['tabs_data_table'], children=[
+                core.Tab(label='Known Barcodes', children=[
+                    dash_table.DataTable(
+                        id=ids["known_data_table"],
+                        columns=KNOWN_DATA_TABLE_COLS,
+                        export_format='csv',
+                        export_headers='names',
+                        sort_action='native',
```
callunity, line 119: For some reason, I'm unable to copy-paste from these tables. I had hoped that adding `include_headers_on_copy_paste=True,` might fix it, but it doesn't seem to. This is merely me expressing sadness rather than requesting you fix Dash, though.

callunity (approved):

application/dash_application/views/bcl2fastq.py
slazicoicr: Yeah, weird bug. The solution is to first Shift-select the text and then copy and paste:

Peek 2020-02-05 13-15

callunity: :crying_cat_face: that's impressively bad.

PR #123, opened 2020-02-06
Single Lane views can only highlight samples that are graphed

slazicoicr:
Makes the drop down box less cluttered and maybe makes initial loading a little faster as the full JSON holding all libraries is no longer sent (big for TS).

This can be extended to Call Ready views when #121 is merged.


callunity (commented):
I see you've got this set up for the single-lane pages; could it be expanded to include the existing call-ready pages as well?

slazicoicr:
Yes. I was waiting for the big changes to be merged into develop, which I see has happened.

callunity (approved):

asofmosq (approved):

PR #125, opened 2020-02-07
[GR-1040] Add "Approve this run in MISO" button to single lane pages

asofmosq:
In order to stop the page from re-loading when a new value was added to runs list, I wrapped the button in a core.Loading but it slows down the page some on initial load. If there's a way to accomplish this without the slowdown, please let me know

### DIFF
--- a/application/dash_application/views/single_lane_rna.py
+++ b/application/dash_application/views/single_lane_rna.py
@@ -23,6 +23,7 @@ title = "Single-Lane RNA-seq"
 ids = init_ids([
     # Buttons
     "update-button",
+    "approve-run-button",
 
     # Sidebar controls
     "all-runs",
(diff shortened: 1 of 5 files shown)


asofmosq:
It's also fastest on the single lane WGS page but considerably slower on the other two

avarsava:
image
Button layout gets squashed when my screen is at 50% width (my usual use case for Dashi), could we use a relative width somewhere to fix this?

avarsava (commented):
This seems more reactive than the rest of the page - if it decided whether to display the button only when Update is pressed, could that fix the issue with needing the separate Loading() for it? Users would probably want to see the graphs for the run before approving it anyway

application/dash_application/views/single_lane_rna.py
```diff
@@ -533,6 +535,17 @@ def update_pressed(click,
...
+    @dash_app.callback(
+        [Output(ids["approve-run-button"], "href"),
+         Output(ids["approve-run-button"], "style")],
+        [Input(ids["run-id-list"], "value")]
+    )
+    def approve_run(value):
+        if len(value) == 1:
+            return ["https://miso.oicr.on.ca/miso/run/alias/" + value[0], {"display": "inline-block"}]
+        else:
+            return ["", {"display": "none"}]
```
callunity: If you wanted, you could move this functionality into sidebar_utils and use one method everywhere.

callunity (approved):

avarsava (approved):
Introduces a tiny display bug when window width is at 50%:
image
But that can be addressed later

callunity (approved):

PR #126, opened 2020-02-10
[GR-926] Change tab title to Dashi

avarsava:
It looks as though changing the <title> per page may be either difficult or impossible. Changing it to 'Dashi' is, however, possible, so this change does that first


avarsava:
Related documentation: https://dash.plot.ly/external-resources

callunity (approved):

apmasell (approved):

PR #127, opened 2020-02-12
GR-846 add call-ready RNA page

callunity:
* can use development ETL rnaseqqc2merged cache to test


avarsava (approved):

asofmosq (approved):

slazicoicr:
`Call-Ready RNA` needs to be added to the home page.

slazicoicr (commented):
* % Correct Read Strand is not informative yet (https://jira.oicr.on.ca/browse/GR-1062) so should be hidden
* There is something off with % rRNA Contamination (most of the samples are at exactly 50%). Should be hidden until that is resolved.

callunity:
I've hidden % correct read strand and % rrna contamination until we can look into them.

application/dash_application/views/call_ready_rna.py
```diff
@@ -0,0 +1,428 @@
...
+
+def get_merged_rna_data():
+    """
+    Join together the RNAseqQC and Pinery dataframes
+    """
+    # Sample metadata from Pinery
+    pinery_samples = util.get_pinery_merged_samples()
+    pinery_samples = util.filter_by_library_design(pinery_samples, util.rna_lib_designs)
+
+    rnaseqqc_df = util.get_rnaseqqc_merged()
```
slazicoicr: This does not match with the columns used `gsiqcetl.column.RnaSeqQc2MergedColumn`.

application/dash_application/views/call_ready_rna.py
```diff
@@ -0,0 +1,428 @@
...
+    rnaseqqc_df = util.filter_by_library_design(rnaseqqc_df, util.rna_lib_designs, RNASEQQC_COL.LibraryDesign)
+
+    rnaseqqc_df[special_cols["Total Reads (Passed Filter)"]] = round(
+        rnaseqqc_df[RNASEQQC_COL.TotalReads] / 1e6, 3)
+    rnaseqqc_df[special_cols["Unique Reads (PF)"]] = round(
+        rnaseqqc_df[RNASEQQC_COL.UniqueReads] / 1e6, 3)
+    rnaseqqc_df[special_cols["% rRNA Contamination"]] = round(
+        (rnaseqqc_df[RNASEQQC_COL.RRnaContaminationInTotal] / rnaseqqc_df[
+            RNASEQQC_COL.TotalReads]) * 100, 2)
```
slazicoicr: If this view will use rnaseqqc (and not rnaseqqc2) then this can be replaced with `rRNAContaminationreadsaligned` (which is used by Single Lane RNA-Seq).

application/dash_application/views/call_ready_rna.py
callunity: I've updated the df_manipulation file to make it clearer that everything for rnaseqqc merged is rnaseqqc2.

application/dash_application/views/call_ready_rna.py
callunity: As this uses rnaseqqc2, we don't have that field available.

slazicoicr (approved):

PR #131, opened 2020-02-13
[GR-1013][GR-1041] Jira Ticket Buttons

asofmosq:
https://jira.oicr.on.ca/browse/GR-1013

https://jira.oicr.on.ca/browse/GR-1041

Where the "file a ticket about these runs" on single-lane pages is visible after the update button is clicked.

### DIFF
--- a/application/dash_application/utility/sidebar_utils.py
+++ b/application/dash_application/utility/sidebar_utils.py
@@ -325,15 +325,28 @@ def jira_button(button_text: str, button_id: str, style: Dict, href: str) -> htm
                   href=href)
 
 
-def construct_jira_link(runs) -> str:
+def construct_jira_link(runs, page_name) -> str:
     base_url = "https://jira.oicr.on.ca/secure/CreateIssueDetails!init.jspa?issuetype=9&pid=11684&priority=10000" \
                "&labels=dashi"
+    summary = {"summary": page_name + ": "}
+    base_url = base_url + "&" + urllib.parse.urlencode(summary)
+
     if runs:
         run_string = "Runs: " + ", ".join(str(run) for run in runs)
         description = {"description": run_string}
         return base_url + "&" + urllib.parse.urlencode(description)
     else:
         return base_url
+
+
+def jira_display_button(runs: List[str], page_title: str):
+    """ Don't display "File a ticket about these runs" button if more than 100 runs are selected.
+     This prevents an HTTP 400 "Request Header Too Large" error """
+    if len(runs) <= 100:
+        return [construct_jira_link(runs, page_title),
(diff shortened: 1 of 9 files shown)


avarsava (commented):
I tried filing a ticket with all the runs selected and got HTTP 400 Bad Request "Request header is too large", which makes sense since it's delivering every run through the URL. Is there anything we could do about this? Squashing all the runs to 'all_runs' like we do for logging would cause a bug report to lose state when the data updates

callunity:
> I tried filing a ticket with all the runs selected and got HTTP 400 Bad Request "Request header is too large", which makes sense since it's delivering every run through the URL. Is there anything we could do about this? Squashing all the runs to 'all_runs' like we do for logging would cause a bug report to lose state when the data updates

I can't imagine a meaningful reason to file a ticket with all runs, so it might make sense to check the URL character count and request that users select fewer runs if it exceeds a certain limit. 
I think 50 would be a reasonable maximum number of runs. So, 50 runs * (29 characters per run name + n separator characters) = max URL size before Dashi requests you choose fewer runs, which should be much less than the Firefox & Chrome URL size limit (~65000 characters) (though we should make sure there isn't a Jira URL size limit).

slazicoicr (commented):
* To save vertical space, would it make sense to have the `Update` and `File Ticket` buttons in the same row?
* I would include the view from which the ticket was filed in the JIRA Summary field: `Dashi RNA-Seq Single Lane: `

asofmosq:
> To save vertical space, would it make sense to have the Update and File Ticket buttons in the same row?

It hasn't been included in this upcoming release but for the single-lane pages, when only a single run is selected an "Approve this run in MISO" button already appears next to the Update button. 

> I would include the view from which the ticket was filed in the JIRA Summary field: Dashi RNA-Seq Single Lane: 

Will update with this amendment.

callunity (approved):

application/dash_application/utility/sidebar_utils.py
```diff
@@ -316,6 +316,39 @@ def parse_query(query) -> List[str]:
...
+        description = {"description": run_string}
+        return base_url + "&" + urllib.parse.urlencode(description)
+    else:
+        return base_url
+
+
+def jira_display_button(runs: List[str], page_title: str):
+    """ Don't display "File a ticket about these runs" button if more than 100 runs are selected.
+     This prevents an HTTP 400 "Request Header Too Large" error """
+    if len(runs) <= 100:
```
avarsava: When 0 runs are selected and 'Update' is pressed, 'File a Ticket About These Runs' still shows up

avarsava (approved):

PR #135, opened 2020-02-18
[GR-932] Subplots to solve WebGL limit in Call Ready TS

slazicoicr:
Subplots seem to successfully get around WebGL limit, so everything was moved into one view again. I have combined four plots into one subplot. Future discussion can be head about how many graphs a subplot should have. 

Subplot should have the following features:
* Legend is shared across all graphs
* X-axis is shared, so zooming in one graph should zoom in the others
* Highlight sample selects the same samples in all graphs


application/dash_application/views/call_ready_ts.py
```diff
@@ -435,26 +479,14 @@ def layout(query_string):
...
-                    # core.Graph(
-                    #     id=ids["at-dropout"],
-                    #     figure=generate_at_dropout(df, initial)),
-                    #
-                    # core.Graph(
-                    #     id=ids["gc-dropout"],
-                    #     figure=generate_gc_dropout(df, initial)),
+                    core.Graph(
+                        id=ids["duplicate-rate"],
+                        figure=generate_duplicate_rate(df, initial)),
```
avarsava, line 464: Why are some graphs included in the multiplot and others aren't?

application/dash_application/views/call_ready_ts.py
```diff
@@ -272,44 +272,91 @@ def generate_duplicate_rate(df, graph_params):
...
-        "%", graph_params["colour_by"], graph_params["shape_by"],
+        graph_params["colour_by"], graph_params["shape_by"],
         graph_params["shownames_val"], [],
-        util.ml_col)
+        util.ml_col,
+        showlegend=False
+    )
+
+
+def generate_subplot(df, graph_params):
```
avarsava: If you passed in an array of generate_*_traces functions and their titles and iterated over them, this could live in plot_builder rather than just this report.

avarsava (commented):
Somewhere along the way the styling for the legends got lost and there's now much more whitespace between items than there was before. I prefer a denser list, but it's OK if there's no fixing it

application/dash_application/views/call_ready_ts.py
avarsava: I'd be willing to discuss a move to using the multiplot for all graphs in all reports after this refactor. I like the sync in zoom between all the plots, however we do lose the ability to save .png files of each graph

slazicoicr:
* Legend whitespace was due to use of `legendgroups`, which allows graphs in a subplot to share a legend. Spacing was set back to 0 in 6f96ada 
* Empty graphs were not correctly displayed in subplots. Fixed in bfa8014 
* Generalized the `generate_subplot` function in cd6bf26

callunity (approved):

slazicoicr:
I am again seeing the WebGL error. Will investigate.

slazicoicr:
After a more skeptical look at my solution, subplots do not solve the WebGL error. Each graph in a subplot counts as 2 WebGL contexts. I will leave this branch alive as an example of how to implement subplots. This still leaves Andre's possible solution:

I did a quick check into the WebGL context thing. There's no official method to force a clean up, but there is an extension to try to clean up a context: https://github.com/pixijs/pixi.js/pull/2481/files
Looking in src/traces/image/plot.js, which I think is where plotly.js does its thing, it seems like it allocates one context per layer and then hopes the garbage collector gets to it. Since it creates a new canvas for each layer, it should be possible to statically allocate one canvas and reuse it. No resource allocation required.

PR #137, opened 2020-02-20
Fixed sorting mismatch and clearer graph names

slazicoicr:


avarsava (approved):

callunity (commented):
Would you mind also making these changes in the commented-out code in `call-ready-ts-1`? I want to avoid a scenario where we uncomment the code in one file and delete the other file, and miss out on some changes that were only present in the (now-deleted) file.
GR-1074

callunity (approved):

PR #139, opened 2020-02-24
All graphs as subplots

callunity:
GR-932
This builds off of @slazicoicr 's previous work and adds only one `core.Graph` element per page, and each individual diagram is implemented as a subplot. 
This PR also slightly shrinks the sidebar and uses the rest of the page's width for graphs.


application/dash_application/views/call_ready_rna.py
```diff
@@ -134,57 +129,64 @@ def get_merged_rna_data():
...
         lambda d: d[util.ml_col],
         lambda d: d[special_cols["% rRNA Contamination"]],
-        "%", graph_params["colour_by"], graph_params["shape_by"],
-        graph_params["shownames_val"],
+        graph_params,
         [(cutoff_rrna_contam_label, graph_params[cutoff_rrna_contam])],
         util.ml_col)
 
 
+graphs = [
```
avarsava: Is there any way to keep the axis labels inside the generate_*_subplot functions? To me this seems like having to look in multiple places for the unchanging parts of each graph, when before everything that wasn't affected by the sidebar was contained in the function

application/dash_application/views/call_ready_rna.py
```diff
@@ -201,7 +203,7 @@ def layout(query_string):
                                           {"display": "inline-block"},
                                           sidebar_utils.construct_jira_link([], title))]),
             html.Div(className="row flex-container", children=[
-                html.Div(className="sidebar four columns", children=[
+                html.Div(className="sidebar three columns", children=[
```
avarsava: Some interesting things happen to the sidebar when the browser window is at 50% width 
image

avarsava (commented):
With this change we lose the 'download plot as png' for each individual graph, and I didn't notice in the previous ticket but the result of pressing it for the new multi graph does not seem useful (legend blurred out for github): 
image
How much do we care about this feature?

slazicoicr:
For Call Ready TS, the Graph/Table tab is broken for me (graphs are at the bottom)

image

slazicoicr (commented):
`Second sort` does not work when you click Update button the first time, but works when you click it a second time. This seems to happen in all views tested.

In some circumstances, Single Lane RNA view, second sorting by % Correct Read Strand does not work.

application/dash_application/views/call_ready_ts.py
```diff
@@ -408,67 +394,25 @@ def layout(query_string):
...
-                            #     id=ids["gc-dropout"],
-                            #     figure=generate_gc_dropout(df, initial)),
-                        ]),
-                        # Tables tab
-                        core.Tab(label="Tables",
+                                id=ids["graphs"],
+                                figure=generate_graphs(df, initial, graphs)
+                            ),
+                        ])
+                    ]),
```
avarsava: Tabs() is closed before the second Tab() is opened

slazicoicr:
Downloading images is important for Dr. Pugh (easy adding to presentations). I found possible fix at https://community.plot.ly/t/save-plot-as-png-sizing-and-positioning/10166/6

```
You can now pass in configuration to dcc.Graph that fixes sizing issues with downloaded png. The following config will make downloaded png same size as on screen.

config = {"toImageButtonOptions": {"width": None, "height": None}}
dcc.Graph(figure=..., config=config)

This was resolved by a fix in the plotly.js lib: https://github.com/plotly/plotly.js/pull/3746 
```

application/dash_application/views/call_ready_rna.py
callunity: (I'm actively working on this because it's a good idea, but it's taking a bit longer than I expected!)

callunity:
> `Second sort` does not work when you click Update button the first time, but works when you click it a second time. This seems to happen in all views tested.

I'm not able to reproduce this... I've been trying with:
  * load page
  * click "All Projects", select a different "Second Sort" value, then click "Update"
And it just works for me, on the first time clicking "Update". Is this different than how you've been testing it?

application/dash_application/views/call_ready_rna.py
callunity: Well that's impressively ugly. I'll update it to four + eight columns which will still give us more space for graphs.

callunity: Updated to use `one-third column` and `two-thirds column`, please let me know how that looks.

(It turns out using `four columns` and `eight columns` looks terrible because there are two sets of `@media (min-width: 550px)` queries, and the later (overriding) one only goes up to `seven columns` so it uses the `four columns` from the later media query and `eight columns` from the earlier query and they calculate element width differently so the combined width was too large for the screen and the tabs overflowed below the sidebar :((((((  )

application/dash_application/views/call_ready_rna.py
callunity: Refactored to add "Subplot" classes which store the graph traces, title, and y_label; let me know what you think!

application/dash_application/views/call_ready_ts.py
callunity: Fixed

callunity:
@slazicoicr @avarsava I've updated the graphs to incorporate that download png sizing issue patch.

avarsava:
> > `Second sort` does not work when you click Update button the first time, but works when you click it a second time. This seems to happen in all views tested.
> 
> I'm not able to reproduce this... I've been trying with:
> 
>     * load page
> 
>     * click "All Projects", select a different "Second Sort" value, then click "Update"
>       And it just works for me, on the first time clicking "Update". Is this different than how you've been testing it?

Single Lane TS:
Select a run, hit 'update':
image

Switch 'Second Sort' to 'Non-Primary Reads', hit 'update':
image

Hit 'Update' again:
image

(all this said, not sure if this PR is the change to cause this behaviour, didn't try it on develop)

application/dash_application/views/call_ready_rna.py
avarsava: Seems better!
image
(The 'Approve this Run in MISO' button still gets cut off awkwardly, but idk if there's any helping that one)

application/dash_application/utility/plot_builder.py
```diff
@@ -309,11 +389,72 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
...
                 }
+            },
+            legend={
+                'tracegroupgap': 0
             }
         )
     )
 
 
+def generate_subplot(subplots: List[Subplot]):
```
avarsava, line 400: I think this was a source of confusion for me - the terminology plotly uses is that one plot holds multiple subplots, it seems like we've been saying one subplot holds multiple subplots? So 'generate_subplot' to me suggests it generates just 1 of the subplots, not all of them

application/dash_application/utility/plot_builder.py
```diff
@@ -290,16 +347,39 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
...
                 "size": highlight_df['markersize'],
                 "opacity": 1
             }
         ))
+
+    return traces
+
+
+# writing a factory may be peak Java poisoning but it might help with all these parameters
+def generate(title_text, sorted_data, x_fn, y_fn, axis_text, graph_params,
```
avarsava, line 364: Could this be renamed to explain better what it's generating? (vs generate_subplot, generate_traces, generate_graphs etc)

application/dash_application/utility/plot_builder.py
```diff
@@ -309,11 +389,72 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
...
+    Generates a subplot using functions that take a DataFrame and graph paramaters,
+    returning a list of traces.
+
+    """
+    fig = make_subplots(
+        rows=len(subplots),
+        cols=1,
+        vertical_spacing=0.02,
+        shared_xaxes=True,
+        subplot_titles=[subplot.title for subplot in subplots]
```
avarsava, line 411: Ah, I see now why the titles couldn't have gone in the generate_*_subplot functions. What a strange decision on plotly's part

avarsava (approved):
I think I've got my head wrapped around it, but I'd still appreciate some more documentation about the flow between all the generate_*() functions, since there's many of them now, and their internal logic

slazicoicr:
Second sort does not work for me properly on Firefox or Chromium (requires update button to be clicked twice), but does work on the current dev branch (70af177). Something about subplots causes this weird behaviour.

slazicoicr:
Code transplanted in #207

PR #142, opened 2020-03-04
Update call-ready WGS

callunity:
GR-1080 & GR-1081

This includes an autoformatting commit which makes it look way more complex than it actually is.

### DIFF
--- a/application/dash_application/views/call_ready_wgs.py
+++ b/application/dash_application/views/call_ready_wgs.py
@@ -56,8 +56,6 @@ ids = init_ids([
     "callability",
     "mean-insert",
     "duplicate-rate",
-    "purity",
-    "ploidy",
     "unmapped-reads",
 
     # Tables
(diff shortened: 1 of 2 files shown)


application/dash_application/views/call_ready_wgs.py
```diff
@@ -212,7 +215,7 @@ def generate_deduplicated_coverage(df, graph_params):
 
 def generate_callability(df, graph_params):
     return generate(
-        "Callability (14x/18x)", df,
+        "Callability (14x/8x) (%)", df,
```
avarsava, line 218: was 18 -> 8 supposed to happen?

apmasell: Yes. This was a mistake.

apmasell (approved):

avarsava (approved):

PR #143, opened 2020-03-04
Improve graph titles, sources, axes

callunity:
GR-1098, GR-1089, GR-1066, GR-1092, GR-1103

### DIFF
--- a/application/dash_application/utility/plot_builder.py
+++ b/application/dash_application/utility/plot_builder.py
@@ -231,6 +231,14 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
                 t=50,
                 pad=4
             )
+    y_axis = {
+        'title': {
+            'text': axis_text
+        }
+    }
+    if axis_text == '%':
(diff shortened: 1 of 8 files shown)


application/dash_application/utility/plot_builder.py
```diff
@@ -402,7 +403,7 @@ def dropdown():
             {"label": "Project", "value": PINERY_COL.StudyTitle},
             {"label": "Run", "value": PINERY_COL.SequencerRunName},
             {"label": "Kit", "value": PINERY_COL.PrepKit},
-            {"label": "Tissue Prep", "value": PINERY_COL.TissuePreparation},
+            {"label": "Tissue Material", "value": PINERY_COL.TissuePreparation},
```
avarsava, line 406: To confirm, everything but the column name should change 'prep' -> 'material'?

avarsava (approved):

application/dash_application/utility/plot_builder.py
callunity: Yes, that's correct

apmasell (approved):

PR #147, opened 2020-03-18
[GR-1086] RNA-Seq 5/3 bias is on log scale

slazicoicr:


application/dash_application/views/call_ready_rna.py
```diff
@@ -135,13 +135,15 @@ def get_merged_rna_data():
...
 def generate_five_to_three(df, graph_params):
-    return generate(
+    fig = generate(
        "5 to 3 Prime Bias", df,
        lambda d: d[util.ml_col],
        lambda d: d[RNASEQQC2_COL.MetricsMedian5PrimeTo3PrimeBias],
        "", graph_params["colour_by"], graph_params["shape_by"],
        graph_params["shownames_val"], [],
        util.ml_col)
+    fig.update_layout(yaxis_type="log")
```
callunity, line 146: Would it make sense to also update the title on line 142 to be "Log Ratio"?

callunity (approved):
Screen Shot 2020-03-18 at 9 54 19 AM
On the current call-ready RNA page there seems to be something funny going on with the y-axis in that it's not immediately obvious to me where `1` is, but it looks like that's a Dash framework issue. So long as this is easily understandable by our users, I think it's good to go.

application/dash_application/views/call_ready_rna.py
slazicoicr: Fixed. The strange y-axis numbering is a design choice by plotly: https://plot.ly/python/axes/#logarithmic-axes.

callunity: 👍 to updated axis title, 👎 to this design choice (given our particular use case scenario, where our data points are around 1).

avarsava (approved):

PR #149, opened 2020-03-19
[GR-1091] Single lane targeted sequencing uses bamqc3

slazicoicr:
I compared data between bamqc and bamqc3 that is displayed in dashi. One is normal scale, the other is log scale to show differences near the low values (MiSeq all clump together).

Normal scales:
plot

Log scales:
plot_log

* Unmapped Reads: Looks good. Log scale shows a couple of outliers in the MiSeq level.
* Non-Primary Reads: bamqc has higher numbers than bamqc3. No idea why, as non-primary reads is fairly straightforward to count. bamqc3 uses `samtools stats`, which I trust more than the homebrew solution in bamqc
* Insert Mean: Different. Again bamqc3 uses `samtools stats`.
* Total Reads: bamqc is consistently higher than bamqc3. Best guess is that bamqc includes non-primary reads. I trust bamqc3, as I have verified it against `samtools flagstat`
* Reads on Target: Looks fairly good, with some MiSeq outlier groups.

### DIFF
--- a/application/dash_application/views/single_lane_ts.py
+++ b/application/dash_application/views/single_lane_ts.py
@@ -9,7 +9,7 @@ from ..utility.table_builder import table_tabs, cutoff_table_data_ius
 from ..utility import df_manipulation as util
 from ..utility import sidebar_utils
 from ..utility import log_utils
-from gsiqcetl.column import BamQcColumn
+from gsiqcetl.column import BamQc3Column
 import pinery
 import logging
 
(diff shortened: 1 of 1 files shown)


callunity (approved):

avarsava (approved):

iainrb:
+1 to not trusting the homebrew bamqc. It's an unspeakable horror of hastily written Perl code.

iainrb (approved):

PR #152, opened 2020-03-25
[GR-1100] Stacked bar chart for Percentage On/Near/Off Bait

avarsava:
Doesn't use our standard colours nor is it very wisely coded, but it does what it needs to. Would like to refactor after the release.

Can change location of chart on request.

### DIFF
--- a/application/dash_application/views/call_ready_ts_2.py
+++ b/application/dash_application/views/call_ready_ts_2.py
@@ -307,17 +307,13 @@ def generate_gc_dropout(df, graph_params):
 
 def generate_bar():
     datas = [[1,1], [1,3]]
-    traces = []
-    for datum in datas:
-        graph = go.Bar(
-            x = [datum[0]],
-            y = [datum[1]]
-        )
-
-        traces.append(graph)
+    graph = go.Bar(
+        x = datas[0],
+        y = datas[1]
+    )
 
     return go.Figure(
-        data = traces
+        data = graph
     )
 
 
--- a/application/dash_application/views/call_ready_ts_2.py
+++ b/application/dash_application/views/call_ready_ts_2.py
@@ -305,6 +305,21 @@ def generate_gc_dropout(df, graph_params):
         graph_params["shownames_val"], [],
         util.ml_col)
 
+def generate_bar():
+    datas = [[1,1], [1,3]]
+    traces = []
+    for datum in datas:
+        graph = go.Bar(
+            x = [datum[0]],
+            y = [datum[1]]
+        )
+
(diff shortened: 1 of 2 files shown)


apmasell (approved):

application/dash_application/views/call_ready_ts_2.py
```diff
@@ -305,6 +309,32 @@ def generate_gc_dropout(df, graph_params):
...
+        graph = go.Bar(
+            name = col + " (%)",
+            x = df[util.ml_col],
+            y = df[col] / df[special_cols["Total Bait Bases"]]
+        )
+        graphs.append(graph)
+    
+    figure = go.Figure(
+        data = graphs,
+        layout = go.Layout(
```
slazicoicr, line 325: Match x-axis and margin look from other graphs: https://github.com/oicr-gsi/dashi/blob/81eb8f5d9b5c9f1488f4b6b76ff055d114e54757/application/dash_application/utility/plot_builder.py#L251-L254

avarsava: Are we sure we want to remove the x-axis labels this time? The legend on the righthand side does not show the names, it explains the colours of the bars. Without the x-axis labels the only way to discover the names is to hover

avarsava: That said, they're not particularly readable, just suggest what the bars are. Working on a compromise with axis labels but it's not going well

slazicoicr: Yeah, it's tricky. In other plots, the legend shows color/first sort and shape. The user has to hover to get names. In the bar, the x-axis show the names, but they get cropped and for most projects (OCT) there are enough bar charts that plotly hides most of the x-axis labels anyways. If we are going to keep the X-axis, I would suggest it just shows the color/first sort value to match the legend from the other graphs.

Given that the bar charts are sorted by color/first sort, would it makes sense to alternate subtle formatting to indicate to the user which bar group is which. Maybe use opacity. So if a user selects only one Project to color by, all bars would be the same opacity. If there are three projects, the first project group has no opacity, the second project group has 10% opacity, and the third project group has no opacity again.

avarsava: There's no grouping on the data without a refactor and i think we're trying to get this feature out for the release today. I can remove the x-axis labels entirely and hope that users understand to hover to get information on the graphed data for now

avarsava: Updated

slazicoicr (approved):

PR #154, opened 2020-03-26
[GR-966] Proof-of-principle genome build features

slazicoicr:
Requires ETL v0.23


avarsava (commented):
edit: nvm missed that that's automatic! Please consider adding a collapsing_function for references though
Differing terminology in the sidebar where it's called 'Genome build' up top but 'Reference' in the dropdowns might be confusing.

avarsava (approved):

morgantaschuk (approved):

PR #156, opened 2020-03-26
[GR-1120] Bottom Update button

avarsava:
Adds an Update button to the bottom of the sidebar that performs the same as the top one


apmasell (approved):
If only this was the solution to the problem where you have to click update twice.

asofmosq (approved):

PR #157, opened 2020-04-03
update pinery url, we will eventually not track this

avarsava:


.docker/start.sh
```diff
@@ -16,7 +16,7 @@ fi
 # mount your private key for bitbucket into the container to allow this command to succeed
 pip install --trusted-host pypi.python.org -r requirements.txt
 
-export PINERY_URL='http://seqbio-pinery-prod-www.hpc.oicr.on.ca:8080/pinery-ws-miso/'
+export PINERY_URL='http://pinery.gsi.oicr.on.ca'
```
djcooke: Not sure what uses this, but might want to add a trailing slash as there was one in the old URL. Might break things otherwise

djcooke (approved):

.docker/start.sh
avarsava: fixed

apmasell (approved):

PR #159, opened 2020-04-07
Gr 932 subplots without shared axis

slazicoicr:
This is a rebase of #139 with the linked X-axis removed to sidestep plotly subplot sorting bug.


slazicoicr:
Taken up in #207

PR #162, opened 2020-04-13
[GR-1136] Data table works in SARS-CoV-2 view

slazicoicr:
Fixes Data Table issue from #160.


apmasell (approved):

avarsava:
'Raw data' view shows a lot of zeroes - is that a problem with Dashi or the data itself?

avarsava (approved):
Please fix the WIP symbol on the 'Tables' tab label

slazicoicr:
Regarding lots of zeros: I think these might be the negative controls and very low virus tests.

PR #163, opened 2020-04-14
[GR-1136] X-Axis Labels for Coverage Percentile graph

avarsava:
(Also, I'm not really sure about that TODO on top of generate_line? Nothing about it is particular to the SARS-CoV-2 report)

### DIFF
--- a/application/dash_application/utility/plot_builder.py
+++ b/application/dash_application/utility/plot_builder.py
@@ -360,7 +360,6 @@ def generate_bar(df, criteria, x_fn, y_fn, title_text, yaxis_text):
     return figure
 
 
-# TODO: Make this more general. Currently it is written for the SARS-CoV-2 view
 def generate_line(df, criteria, x_fn, y_fn, title_text, yaxis_text, xaxis_text=None):
     graphs = []
     for name, df in df.groupby(criteria):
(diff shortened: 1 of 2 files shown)


application/dash_application/utility/plot_builder.py
```diff
@@ -361,7 +361,7 @@ def generate_bar(df, criteria, x_fn, y_fn, title_text, yaxis_text):
 
 
 # TODO: Make this more general. Currently it is written for the SARS-CoV-2 view
-def generate_line(df, criteria, x_fn, y_fn, title_text, yaxis_text):
+def generate_line(df, criteria, x_fn, y_fn, title_text, yaxis_text, xaxis_text=""):
```
slazicoicr: I would make this `xaxis_text=None` and having `'visible': xaxis_text is not None`. This way one could still show the x-axis values and not have a x-axis title.

Regarding the TODO. You can remove it. I mean it did not have all the features of the scatter graph (shape, cutoff lines, etc), but there is no need to make unnecessary work.

slazicoicr (approved):

application/dash_application/utility/plot_builder.py
avarsava: Both fixed

apmasell (approved):

PR #166, opened 2020-04-15
Fixed bugs due to use of raw string (poped up after update to ETL 0.27)

slazicoicr:


application/dash_application/utility/df_manipulation.py
```diff
@@ -158,19 +158,37 @@ def normalized_merged(df: DataFrame, merged_cols: List[str]):
...
+    PINERY_COL.ParentSampleName,
+    PINERY_COL.TemplateType,
+    PINERY_COL.RunBaseMask,
+    PINERY_COL.RunDir,
+    PINERY_COL.LastModified,
+    PINERY_COL.SequencerRunPlatformModel,
+    PINERY_COL.ReceiveDate,
+    PINERY_COL.TissueRegion,
+    PINERY_COL.WorkflowType,
+    PINERY_COL.Skip,
```
avarsava, line 190: 44 columns named previously, 24 here - do we no longer need to filter those columns out?

slazicoicr: Yeah, I could not quite understand why the original version worked. It is removing mandatory fields (`TissuePreparation`, `SampleName`), which should have caused a crashed Dashi. `drop` would have raised an exception if any of the columns to be dropped were missing (column names were different), so they were definitely present.

The way I build my list is to start out with all columns from the previous list and slowly delete the ones that cause Dashi to crash.

apmasell: The bug in ETL caused it to inject columns with NaNs.

slazicoicr: The remaining columns after the `drop` call:
`['studyTitle', 'rootSampleName', 'sampleName', 'sequencerRunName', 'laneNumber', 'iusTag', 'instrument_name', 'geo_external_name', 'geo_library_source_template_type', 'geo_organism', 'geo_prep_kit', 'geo_tissue_origin', 'geo_tissue_type', 'umis', 'subproject', 'geo_targeted_resequencing', 'geo_tissue_preparation', 'rin', 'institute', 'geo_group_id', 'dv200', 'sequencing_control_type', 'Sample Type']`

apmasell (approved):

avarsava (approved):

PR #168, opened 2020-04-16
[GR-1136] Active Projects issue seems to have resolved on its own

avarsava:


apmasell (approved):

application/dash_application/utility/df_manipulation.py
```diff
@@ -6,7 +6,7 @@
 from gsiqcetl import QCETLCache
 import gsiqcetl.column
 import pinery
-
+import pdb
```
slazicoicr: remove debug import

slazicoicr (approved):

application/dash_application/utility/df_manipulation.py
avarsava: done

PR #170, opened 2020-04-17
[GR-1151] Use median coverage with 10/90 percentile bars

slazicoicr:


apmasell (approved):

avarsava:
Some shape/colour/sort combinations cause points to escape their bars, one of probably many is:
Sort: Project
2nd Sort: SARS-CoV-2 Percentage
Colour By: Run
Shape By: Sequencing Control Type 
image

slazicoicr:
This seems to be a bug with plotly. Changing to `Scatter` instead of `Scattergl` leads to correct display (but the error bars are black).

image

There are several open issues in plotly that might fix this:
https://github.com/plotly/plotly.js/issues/2900
https://github.com/plotly/plotly.js/issues/2882

avarsava:
Just tried it out. I'm OK with the black bars until the bug with Scattergl is fixed, however if we go with this I'd like a `generate_with_error_bars` or something so that only this graph uses svg and the rest still use WebGL

slazicoicr:
I got colours to work. The approach was to use `Scatter` if error bars are used, otherwise use `Scattergl`.

avarsava (approved):

PR #172, opened 2020-04-20
[GR-1136] Move caches to df_manipulation

avarsava:


avarsava:
There's multiple caches in use - which should I use for DATAVERSION?

apmasell:
How did we end up with multiple caches?

apmasell (approved):

slazicoicr:
The Covid view pulls data from multiple caches (bedtools, samtools, kraken2).

avarsava:
thanks @slazicoicr , andre helped me out on that and i pushed changes to include the dataversions

slazicoicr (approved):

PR #174, opened 2020-04-20
[GR-1152] Covid on target bar chart no longer shows unmapped numbers

slazicoicr:


apmasell (approved):

avarsava (commented):
does `special_columns` have a purpose beyond this chart? Do we need it in the data processing section after this change?

slazicoicr:
It still has a use in the data processing section (gets around having to write out those columns multiple times).

avarsava (approved):

PR #185, opened 2020-05-01
[GR-1158] Show only On/Near Bases on TS Bar graph

slazicoicr:


apmasell (approved):

application/dash_application/views/call_ready_ts_2.py
```diff
@@ -317,50 +317,16 @@ def generate_gc_dropout(df, graph_params):
...
-                r=50,
-                b=50,
-                t=50,
-                pad=4
-            )
-        )
-    )
-    figure.update_layout(barmode='stack')
-
-    return figure
```
avarsava, line 363: :facepalm: thanks for cleaning this up

avarsava (approved):

PR #193, opened 2020-05-13
[GR-1168] Always autoscale y-axis

avarsava:
Undoes previous decision to always lock y-axis to 0-100 when graphing percentage values, as discussed in QC Meeting 12/05/2020. 
Allows for automatic zoom in cases like mentioned in GR-1168. 
Also reveals that some runs have percentage values >100%. 
Requesting that reviewers launch Dashi with this change as it's a pretty significant change to Dashi's UX


apmasell (approved):

iainrb (approved):

slazicoicr (approved):
Seeing > 100% is for the best, as that's a bug either is the workflow or the data processing.

PR #194, opened 2020-05-14
[GR-1082] Use median insert size with 10/90 percentiles

slazicoicr:
Does not include median coverage, as that is trickier to calculate than previously thought (https://jira.oicr.on.ca/browse/GR-1170)


apmasell (approved):

avarsava (approved):
Please update changelog and 'what is dashi graphing' wiki page

PR #196, opened 2020-05-19
Remove Purity from Call-Ready TS 2

avarsava:
why does the changelog portion of the release procedure never work right


avarsava:
Also, does this mean I can remove all of the ichorcna processing from the report or is that data still needed?

asofmosq (approved):

apmasell (approved):

PR #198, opened 2020-05-21
[GR-1154] Big tables for home page

avarsava:
Serverside for speed. Jinja hurts to write. Pretty can come later. Open to suggestions on default sort


apmasell (approved):
It makes me nostalgic for Netscape.

asofmosq (approved):

PR #202, opened 2020-05-26
[GR-1171] Remove FastQC dependancy from Single Lane RNA

slazicoicr:


avarsava (approved):
Please also update the changelog and the wiki

apmasell (approved):

PR #203, opened 2020-05-26
[GR-1176] Make date range picker clearable

avarsava:
with bonus css nonsense


apmasell (approved):

slazicoicr (approved):

avarsava:
Force pushed to undo git mistake, don't want to take credit for savo's work

PR #204, opened 2020-05-27
Seperate trace creation from Figure creation (makes subplots possible)

slazicoicr:
Rebasing #159 turned into a Gordian knot, so I am doing gradual feature roll out to support subplots.

### DIFF
--- a/application/dash_application/utility/plot_builder.py
+++ b/application/dash_application/utility/plot_builder.py
@@ -241,18 +241,18 @@ def reshape_call_ready_df(df, projects, references, tissue_preps, sample_types,
     df = fill_in_size_col(df, searchsample, True)
     return df
 
+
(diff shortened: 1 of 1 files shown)


application/dash_application/utility/plot_builder.py
```diff
@@ -270,28 +270,63 @@ def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
...
+                   'autorange': True},
+            yaxis=y_axis,
+            legend = {
+                'tracegroupgap': 0,
+            },
+        )
+    )
+
+
+def generate_traces(
```
avarsava: Do you intend for reports to call generate_traces() directly as well as calling generate()? If not, please rename to _generate_traces() to match _define_graph()

avarsava (approved):

apmasell (approved):

PR #207, opened 2020-05-28
Subplot infrastructure in anticipation of sorting axis Plotly fix

slazicoicr:
Subplot infrastructure. Only Single Lane RNA-Seq has been converted to subplot for a cautious rollout.


application/dash_application/utility/plot_builder.py
```diff
@@ -1,7 +1,9 @@
-from typing import List, Tuple, Union, Dict
+from typing import List, Tuple, Union, Dict, Callable
 
 import pandas
 import plotly.graph_objects as go
+from plotly.subplots import make_subplots
+import dash_core_components
```
avarsava: Please import as 'core'

application/dash_application/utility/plot_builder.py
```diff
@@ -769,4 +772,247 @@ def items_for_df(self):
...
+            title,
+            df,
+            x_col,
+            y_col,
+            y_label,
+            colourby,
+            shapeby,
+            hovertext_cols,
+            cutoff_lines=None,
+            name_col=PINERY_COL.SampleName,
```
avarsava: Is this true for Call Ready pages? I thought we used RootSampleName or ml_col. Otherwise these constructors seem identical?

application/dash_application/utility/plot_builder.py
```diff
@@ -769,4 +772,247 @@ def items_for_df(self):
...
+    UNIQUE_READS = "🚧 Unique Reads (Passed Filter) (%) -- DATA MAY BE SUSPECT 🚧"
+    UNMAPPED_READS = "Unmapped Reads (%)"
+    UNMAPPED_READS_COUNTS = "Unmapped Reads"
+    X = "x"
+
+
+def generate_subplot(subplots: List[Subplot]):
+    """
+    Generates a subplot using functions that take a DataFrame and graph paramaters,
+    returning a list of traces.
```
avarsava: I'm going to pick on this naming convention again - 'generate subplot'(singular) that takes multiple 'subplots' and is really drawing multiple 'subplots' (all contained in one singular 'subplot'? or 'subplot graph'?) seems like it's going to get confusing fast. Plotly's convention is that one 'plot' contains multiple 'subplots'. Could we rename these to 'generate_subplots[_from_func]' and change create_subplot_graph to either 'create_plot[_with_subplots?]' or else adopt a new term 'superplot' or something for clarity?

application/dash_application/utility/plot_builder.py
```diff
@@ -769,4 +772,247 @@ def items_for_df(self):
...
+            self.hovertext_cols,
+            self.cutoff_lines,
+            self.name_col,
+            self.markermode,
+            self.bar_positive,
+            self.bar_negative,
+        )
+
+    def is_pct_graph(self):
+        return self.y_label == GraphTitles.PCT
```
avarsava: this was previously necessary because we'd autoscale everything but the pct graphs, which were stuck at 0-100. Now that we're autoscaling everything again, is this check needed?

application/dash_application/utility/plot_builder.py
```diff
@@ -769,4 +772,247 @@ def items_for_df(self):
...
+            cutoff_lines,
+            name_col,
+            markermode,
+            bar_positive,
+            bar_negative,
+            log_y,
+        )
+
+
+class GraphTitles:
```
avarsava: This could probably inherit Enum if you wanted

application/dash_application/utility/plot_builder.py
slazicoicr: Hmm, there seems to be only one place `name_col` is used: 

https://github.com/oicr-gsi/dashi/blob/8b453a9cf24eba57ed659a701f43083396b6e3e4/application/dash_application/utility/plot_builder.py#L347-L356

Should this be replaced by `x=x_fn(data)` to match the other x axis coordinates:
https://github.com/oicr-gsi/dashi/blob/8b453a9cf24eba57ed659a701f43083396b6e3e4/application/dash_application/utility/plot_builder.py#L362
https://github.com/oicr-gsi/dashi/blob/8b453a9cf24eba57ed659a701f43083396b6e3e4/application/dash_application/utility/plot_builder.py#L377
https://github.com/oicr-gsi/dashi/blob/8b453a9cf24eba57ed659a701f43083396b6e3e4/application/dash_application/utility/plot_builder.py#L541

application/dash_application/utility/plot_builder.py
slazicoicr: `generate_subplot` -> `generate_plot_with_subplots`
`create_subplot_graph` -> `create_graph_element_with_subplots`

Let me know if these makes sense.

application/dash_application/utility/plot_builder.py
slazicoicr: It's a utility function that is never used. I will delete it.

application/dash_application/utility/plot_builder.py
slazicoicr: Enum would make more sense. These are currently not used. I think Heather's idea was to have a central repository of graph titles. i am not sure if that needs to be done or if this is the best way.

application/dash_application/utility/plot_builder.py
avarsava: That sounds great, thanks

application/dash_application/utility/plot_builder.py
avarsava: Seems to me like it's not really solving any problem - we once wanted to rename all the Total Reads graphs but we've since done that and it's all been fine, we haven't needed to rename en masse

application/dash_application/utility/plot_builder.py
avarsava: name_col originated, I believe, in the early days when there were only Single-Lane graphs, because x was _always_ just PineryColumn.SampleName, so that was an easy way to not have to specify that for the x axis on every graph definition. I'd like for that idea to come back - only specifying the x-axis function if it's not the Single-Lane or Call-Ready default x-axis

slazicoicr: I will do a PR to remove `name_col` for now then as it would cause a bug if `x_fn` was different from `name_col`. Setting a default for `x_fn` is more annoying, as it would require changing the signature of all views that call `generate` (let's leave that for a broader refactor).

application/dash_application/utility/plot_builder.py
slazicoicr: Left a TODO. I think this feeds into Dashi giving glossary information, but that is a whole bigger fish.

application/dash_application/utility/plot_builder.py
slazicoicr: See #209

apmasell (approved):

application/dash_application/utility/plot_builder.py
```diff
@@ -791,4 +794,241 @@ def items_for_df(self):
...
+            cutoff_lines,
+            markermode,
+            bar_positive,
+            bar_negative,
+            log_y,
+        )
+
+
+# TODO: Currently not in use. Big question: How to deal with titles in Dashi?
+class GraphTitles:
```
apmasell: I would prefer if you deleted this and we can bring it back when we're ready to use it.

asofmosq (approved):

PR #210, opened 2020-05-29
[GR-1187] Remove on target percentage from SARS-CoV-2

slazicoicr:
Graph is misleading and needs to be removed.


slazicoicr:
Might be a way to add proper graphs.

PR #217, opened 2020-06-05
[GR-1181] Single-Lane Date Sort

avarsava:
- Only on Single-Lane TS for now
- the pdb breakpoint (_set_trace()_) in plot_builder is for demonstrating that the dataframe does indeed sort by date after the sort is run. Comment out the if and the breakpoint if you don't want to use pdb
- Available as both First and Second sort because I didn't know which would be better
- Available as Data Label for verifying that no sort happens in front end

### DIFF
--- a/application/dash_application/utility/df_manipulation.py
+++ b/application/dash_application/utility/df_manipulation.py
@@ -383,6 +383,8 @@ def get_pinery_merged_samples(active_projects_only=True):
 def get_runs():
     return _runs_with_instruments.copy(deep=True)
 
+def df_with_runs(df):
+    return pandas.merge(df, get_runs(), left_on=PINERY_COL.SequencerRunName, right_on='name')
 
 def df_with_pinery_samples_ius(df: DataFrame, pinery_samples: DataFrame, ius_cols:
                            List[str]):
--- a/application/dash_application/utility/plot_builder.py
+++ b/application/dash_application/utility/plot_builder.py
@@ -10,6 +10,7 @@ import gsiqcetl.column
 from .df_manipulation import sample_type_col
 from .sidebar_utils import runs_in_range
 import re
+import pdb
 
 PINERY_COL = pinery.column.SampleProvenanceColumn
 COMMON_COL = gsiqcetl.column.ColumnNames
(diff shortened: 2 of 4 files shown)


slazicoicr:
Hmm, playing around with this, there might be a deeper sorting issue going on (shudder). Below is the buggy situation. Reference (sort), Total Reads (second sort), and Run (colour by). The bug is that the data points are sorted first by Run and not Reference.

image

If I change to Reference (colour by), it looks right:

image

I have a vague memory of Heather fighting this sorting issue. I see that for Call Ready WGS, the sort and colour by are always the same. Using git blame leads me to #132.

avarsava:
Yeah, iirc it's a side effect of needing to group data by shape- and colour-by in order to render the legend correctly. I also wrestled with it for a long time a few months ago. I don't think users have complained about it, it seems like it might make intuitive sense even if it doesn't make sense after thinking about it for a second.

slazicoicr:
I think the lack of sorting you see might be because of this. As you cannot colour by date, setting it as first sort will not work.

Second sort by date works. If I sort and colour by Run and second sort by date I see expected behaviour. I think we should link sort and colour by to make this explicit:

image

slazicoicr:
My approach from https://github.com/oicr-gsi/dashi/pull/217#issuecomment-639661108 was cheating, as I was sorting it by Run (which is the date). However, date sorting still works properly when I change sort and colour by to Project. By showing the run when I hover, i can see all points are properly sorted.

avarsava:
Rebased onto develop, now that the sorting fix is in. It's really weird - on Single Lane TS select project BFS, sort by start date, show data label start date. It's SO CLOSE to being right, except that there's a bunch of stray 2018s in with the 2016s. I'll try to figure out why that's happening.

avarsava:
Talked with @slazicoicr. He will investigate some spinoff work on looking at how the x-axis is constructed - Currently we are limited because the x-axis is just library names, so if one library has been run multiple times, all 3 dots get put on the x-axis value for their library at the earliest date. Investigating a composite x-axis in order to properly sort purely by date - if library name clustering is needed, library name sort option will be used.

PR #219, opened 2020-06-09
[GR-1168] Proper axis on empty graphs

slazicoicr:


application/dash_application/utility/plot_builder.py
```diff
@@ -242,6 +242,36 @@ def reshape_call_ready_df(df, projects, references, tissue_preps, sample_types,
...
+
+    Returns:
+
+    """
+    for t in trace_list:
+        try:
+            if t['x']:
+                return False
+        # Pandas Series has: The truth value of a Series is ambiguous.
+        except ValueError:
```
apmasell: It would be a lot nice if this could check if the value is an instance of `Series` rather than catching an exception that might have other reasons for existing.

avarsava (approved):

application/dash_application/utility/plot_builder.py
slazicoicr: Thanks for the push. I found a cleaner way of doing this. The hard thing about duck typing is that a duck can quack in so many different and overlapping ways.

apmasell (approved):
https://www.youtube.com/watch?v=nHc288IPFzk

PR #225, opened 2020-06-16
[GR-1189] bcl2fastq -> bcl2barcode report

avarsava:
- Should anything be done about the Lane Clusters PF column? no such info in bcl2barcode table
- Do I need to filter unknown_run the same as known_run (see commented out code)? Pinery data not available for those runs
- Waiting on ETL update for calculating textarea fraction


avarsava:
Also, could the strexpand file go to a shared location so it doesn't exist in both shesmu and dashi? (Esp if it ever needs to be updated?)

application/dash_application/views/bcl2fastq.py
```diff
@@ -158,41 +176,23 @@ def update_layout(run_alias):
             functions update_known_index_bar, update_unknown_index_bar,
             update_pie_chart's data value, and update_pie_chart's fraction value
         """
+        known_run = known_data_table[known_data_table[bcl2barcode_col.Run] == run_alias]
+        known_run = known_run[~known_run[PINERY_COL.SampleProvenanceID].isna()]
+        known_run = known_run.drop_duplicates([PINERY_COL.SampleProvenanceID, bcl2barcode_col.Lane])
```
apmasell: I don't think this is right. I think you have to group and sum the duplicates. If it duplicated, then there will be multiple barcodes, which are unique entities from the bcl2fastq perspective.

apmasell (commented):
If you can make the strexpand file configurable in some way, we can have it read from `spb-seqware-production` which should already be on those machines from the cron jobs.

application/dash_application/views/bcl2fastq.py
avarsava: This line is only considering the SampleProvenanceID and the Lane number for identifying whether a row is a duplicate. As long as those are duplicated they'll be dropped. Should I also be considering the indices as criteria so they don't get dropped?

apmasell: I'm not sure that makes sense either. I think dropping is the wrong operation. Suppose you have a 10X sample `foo` in lane 1 with barcode `SI-GA-H01`, after joining:

| sampe | lane | pinery barcode | bcl2barcode barcode | count|
|---|---|---|---|---|
| foo | 1 | SI-GA-H01 | AAAA | 32 |
| foo | 1 | SI-GA-H01 | AATT | 30 |
| foo | 1 | SI-GA-H01 | AACC | 33 |
| foo | 1 | SI-GA-H01 | AAGG | 31 |

You want to end up with output that either has 4 stacked bars for `foo`, so, no dropping at all, or a single bar whose height is 32+30+33+31. Does that make sense?

avarsava: Right. I guess I'm confused, then, as to why the original version of this report did it: https://github.com/oicr-gsi/dashi/blob/5ac70e6e53780226402615ed3dbec8829a1a5670/application/dash_application/views/bcl2fastq.py#L165

apmasell: Probably because it was getting duplicate records from the messy data it was ingesting.

avarsava:
> If you can make the strexpand file configurable in some way, we can have it read from `spb-seqware-production` which should already be on those machines from the cron jobs.

Updated

application/dash_application/views/bcl2fastq.py
avarsava: Updated

README.md
```diff
@@ -26,7 +26,7 @@ Create a `.env` file in the root directory of this repository:
 | `LOG_FILE_LOCATION` | **Yes** | File path where logs should be written | `~/logs/dashi.log` | `./dashi.log` |
 | `LOG_TO_CONSOLE` | No | Set to log to console as well as to log file specified above | `True` | do not log |
 | `USE_BLEEDING_EDGE_ETL` | No | Set to install `gsi-qc-etl@master` instead of the release version of `gsi-qc-etl` in `requirements.txt` (Docker only) | `1` | use release version |
-
+| `10X_STREXPAND` | **Yes** | Tab-separated file listing 10X barcodes and 4 sequences for each | `~/barcodes` | |
```
apmasell: Environment variables can't start with numbers. Maybe just `BARCODES_STREXPAND`

apmasell (approved):

README.md
avarsava: Weird that it worked for me, but sure

avarsava: Fixed

application/dash_application/views/bcl2barcode.py
```diff
@@ -28,47 +30,69 @@
     ]
 )
 
+def maybe_index2(s):
+    try:
+        return s.split("-")[1]
+    except IndexError:
+        return ""
+
```
slazicoicr, line 32: Parroting Andre's philosophy of only using Exceptions only in exceptional cases: 

`return "" if len(s.split("-")) == 1 else s.split("-")[1]`

application/dash_application/views/bcl2barcode.py
```diff
@@ -28,47 +30,69 @@
...
+bcl2barcode_col = gsiqcetl.column.Bcl2BarcodeColumn
+pinery = util.get_pinery_samples()
+PINERY_COL = util.PINERY_COL
+
+barcode_expansions = pandas.DataFrame(columns=['Index', 'Sequence'])
+with open(os.getenv("BARCODES_STREXPAND"), 'r') as strexpand:
+    for l in strexpand.readlines():
+        x = l.split("\t")
+        for i in range(1,5):
+            barcode_expansions = barcode_expansions.append({'Index': x[0], 'Sequence': x[i].replace("\n", "")}, ignore_index=True)
```
slazicoicr: `pandas` has the `melt` function to turn a wide DataFrame into a long one.
```
df = pandas.read_csv(
    os.getenv("BARCODES_STREXPAND"),
    sep="\t",
    header=None
)
# The headers are 0, 1, 2, ... (would need to be renamed)
df.melt(id_vars=0)
```

application/dash_application/views/bcl2barcode.py
```diff
@@ -28,47 +30,69 @@
...
+unknown_data_table = bcl2barcode_with_pinery.loc[bcl2barcode_with_pinery['studyTitle'].isnull()]
+
+# Rows which merged successfully are the 'Known' table
+known_data_table = bcl2barcode_with_pinery.loc[bcl2barcode_with_pinery['studyTitle'].notnull()]
 
-unknown = util.get_bcl2fastq_unknown()
-un_col = gsiqcetl.column.Bcl2FastqUnknownColumn
+# Get Known indices from pinery, since it's been expanded properly
+known_data_table['Index1'] = known_data_table['Sequence'].apply(lambda s: s.split("-")[0])
+known_data_table['Index2'] = known_data_table['Sequence'].apply(lambda s: maybe_index2(s))
```
slazicoicr: This warning pops up several times. It occurs first here: `A value is trying to be set on a copy of a slice from a DataFrame`. I think using `copy()` when creating `known_data_table` and `unknown_data_table` should solve it.
`

application/dash_application/views/bcl2barcode.py
```diff
@@ -28,47 +30,69 @@
...
+pinery_with_expanded_barcodes['Sequence'] = pinery_with_expanded_barcodes['Sequence'].fillna(pinery_with_expanded_barcodes['iusTag'])
+
+# Merge the expanded pinery with bcl2barcode to expand the bcl2barcode data the same way
+bcl2barcode_with_pinery = pandas.merge(bcl2barcode, pinery_with_expanded_barcodes, left_on=['Run Alias', 'Lane Number', 'Barcodes'], right_on=['sequencerRunName', 'laneNumber', 'Sequence'], how='left')
+
+# Failures to merge with the pinery data populate the 'Unknown' table
+unknown_data_table = bcl2barcode_with_pinery.loc[bcl2barcode_with_pinery['studyTitle'].isnull()]
+
+# Rows which merged successfully are the 'Known' table
+known_data_table = bcl2barcode_with_pinery.loc[bcl2barcode_with_pinery['studyTitle'].notnull()]
```
slazicoicr: When I look at Run 200504_A00469_0101_AH7MC7DSXY the 10X runs (GNLR) are not shown. They seem to go into the Unknown table.

application/dash_application/views/bcl2barcode.py
```diff
@@ -204,65 +202,59 @@ def create_known_index_bar(run):
...
-            "xaxis": {"title": COL_LIBRARY, "automargin": True},
+            "xaxis": {"title": "Library", "automargin": True},
             "yaxis": {"title": "Clusters"},
+            "showlegend": True
         },
     }
 
 
-def create_unknown_index_bar(pruned):
+def create_unknown_index_bar(run):
```
slazicoicr, line 273: Two suggested user improvements:

* Order index by number of unknown reads (those indices with the most unknown should be first)
* No need to show this many unknown indices. I think bcl2fastq showed the top 20.

application/dash_application/views/bcl2barcode.py
avarsava: I took csvs of both data sources (which I can provide) and compared data for this run on Lane 4 (where all the GNLR info is). Everything in bcl2barcode for this run on that lane has 16-char barcodes. Since the expanded barcodes that go into the pinery df are 8-char, they fail to merge when creating bcl2barcode_with_pinery. 
Do the barcode expansions look like they're being established properly? Is merging on Barcodes and Sequence logical?

apmasell: Right. This is a general problem where the barcodes in Pinery OR in the run, could be shorter. I was wrong in recommending `startswith`. The comparison should be:
```
pinery[0:min(len(pinery), len(bcl2barcode)] == bcl2barcode[0:min(len(pinery), len(bcl2barcode)]
```

I'm not sure how that translates to pandas...

application/dash_application/views/bcl2barcode.py
avarsava: Done

application/dash_application/views/bcl2barcode.py
avarsava: Pushed attempts at implementing this, doesn't seem to work, Savo and I will talk about it this afternoon

slazicoicr:
My approach is to sperate Index1/Index2, find the minimum index length for each run/lane, trim the Index1/Index2 to that minimum length (which allows for merging on them), and then identify known/unknown for both single index libraries and dual index libraries. In the initial approach, there is the big merge on run/lane and then index is compared and trimmed, whereas here I am trimming first to allow for a smaller merge.

application/dash_application/views/bcl2barcode.py
```diff
@@ -57,33 +50,84 @@ def barcode_test_test(row):
...
+pinery_with_expanded_barcodes['Index1'] = pinery_with_expanded_barcodes['Sequence'].fillna(pinery_with_expanded_barcodes['iusTag'])
+pinery_with_expanded_barcodes['Index2'] = pinery_with_expanded_barcodes['Index1'].apply(lambda s: numpy.nan if len(s.split("-")) == 1 else s.split("-")[1])
+pinery_with_expanded_barcodes['Index1'] = pinery_with_expanded_barcodes['Index1'].apply(lambda s: s.split("-")[0])
+pinery_with_expanded_barcodes.loc[pinery_with_expanded_barcodes['Index1'] == 'NoIndex', 'Index1'] = numpy.nan
+
+# Remove Pinery records that don't have Index 1 or Index 2
+pinery_with_expanded_barcodes = pinery_with_expanded_barcodes[~(pinery_with_expanded_barcodes['Index1'].isna() & pinery_with_expanded_barcodes['Index1'].isna())]
+
+# Calculate mininum index length per run/lane
+min_index1_length = pinery_with_expanded_barcodes.groupby(['sequencerRunName', 'laneNumber'])['Index1'].apply(lambda x: x.str.len().min()).reset_index().rename(columns={'Index1': 'Index1 min length'})
```
apmasell, line 65: This has to be per sample. The lab has routine run things that are like:

CCCC
GGGGAT
GGGGCG

so truncating to 4 is not acceptable.

slazicoicr: We are in trade-off land and I am looking for consensus in approach.

In the per sample approach (b9e969e), it is necessary to merge `pinery` and `bcl2barcode` on lane/run, which already results in a huge explosion of rows (2 million) and will get big fast as `pinery` and `bcl2barcode` grows. Then in the huge DataFrame check if single/dual index, trim, and compare barcode for each row (and pandas is gonna be really slow at that as trimming cannot be vectorized).

My approach has the weakness you describe, but I think it can be fixed. The `known` DataFrame can be inspected to check for duplicated barcode/lane/run (`sum(known_data_table.duplicated(['Barcodes', 'Lane Number', 'Run Alias']))`. Currently, there are 0 such duplicates, which makes me a little suspicious of this approach. In future, the first step is to add a warning to Dashi view stating that run has possible index collision and then be even better and smartly check if it is an actual index duplication or a false positive due to my trimming approach.

slazicoicr:
@avarsava Does this look as a good first pass to release? Is anything missing? I talked with Andre and we should definitely have a detailed look at all the gotchas  in the future.

slazicoicr (approved):

avarsava:
@slazicoicr I was waiting for a resolution on the discussion on the per-sample approach. What's the status on that? The report itself looks fine, I can see that GNLR ends up 'known' in the aforementioned run now. I do not know enough about the tool to catch any other edge cases.

slazicoicr:
I think @apmasell approved (I asked for a second approval and it seems to be there). I suggested to Andre that we do a code review of this view to discuss the various edge cases after this PR is in.

avarsava:
I only see a review from Andre 8 days ago. Should we schedule a code review meeting for this PR? I do not want it going out with misleading information.

apmasell:
Yeah, I've been following along. Given this is a work in progress anyway, I think go for it, and we can go through another round.

avarsava:
Should we put a 'work in progress' label on it then? AFAIK the case you describe where barcodes get matched on 4 characters has not been addressed, and unless it's very obvious in the report when that error happens (I can't tell from looking) then I don't want to implicitly rubber stamp this report and have lab decisions be based on it.

apmasell:
Couldn't hurt.

avarsava:
Done. There is still the issue of the 'Predicted / Actual' cluster count textarea, is that count available through ETL now?

slazicoicr:
That text box can be removed. I had it there as bcl2fastq was run multiple times on one lane, which produced its own set of nightmares (clusters being counted multiple times). As bcl2count is only run once, there is no reason to deal with predicted clusters. ETL exposes total clusters for each lane (`QCETLCache().bcl2barcode.run_summary["total clusters"]`), which should probably be used at https://github.com/oicr-gsi/dashi/blob/GR-1189_bcl2barcode/application/dash_application/views/bcl2barcode.py#L323 at QCETL does not return the long tail of barcodes (https://jira.oicr.on.ca/browse/GR-1201).

Let me know if I can help with either fix.

avarsava:
@slazicoicr both addressed, look good?

slazicoicr:
Yes, merge away.

PR #229, opened 2020-06-25
[GR-1205] Add Home button to navbar on reports

avarsava:
Please test to make sure this looks OK on your display! I only have my laptop to test on. It looks good at a variety of window sizes for me.

### DIFF
--- a/application/dash_application/known_pages_router.py
+++ b/application/dash_application/known_pages_router.py
@@ -31,6 +31,8 @@ def navbar(current):
     return dbc.NavbarSimple(
         id="navbar",
         children=[
+            # Simple Navbar only lets you add content to the right side. This CSS forces the button out of the div 
+            html.A(html.Button("Home", style={"position": "fixed", "left": "0", "margin-left": "15px", "margin-top": "4px"}), href="/.."),
             dbc.DropdownMenu(
                 children=[menu_item(v.title, k) for k, v in pages_info.items()],
                 nav=True,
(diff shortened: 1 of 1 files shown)


slazicoicr (approved):
The home button seems to disappear when i make the window narrow (~1/4 of my screen). Don't think this is an issue as the graphs themselves become too narrow at this width.

image

avarsava:
> The home button seems to disappear when i make the window narrow (~1/4 of my screen). Don't think this is an issue as the graphs themselves become too narrow at this width.
> 

Yeah I think I'm OK saying 'there is a width at which Dashi becomes unusably narrow' and it seems to be 1/5th of my screen width

application/dash_application/known_pages_router.py
```diff
@@ -31,6 +31,8 @@ def menu_item(label, link):
     return dbc.NavbarSimple(
         id="navbar",
         children=[
+            # Simple Navbar only lets you add content to the right side. This CSS forces the button out of the div 
+            html.A(html.Button("Home", style={"position": "fixed", "left": "0", "margin-left": "15px", "margin-top": "4px"}), href="/.."),
```
apmasell: I think either `"/"` or `".."`, but both is not doing anything.

apmasell (approved):

application/dash_application/known_pages_router.py
avarsava: Well that's weird, because it works for me. What's your environment like? I'm on firefox 76

apmasell: Oh, it works, it's just _go to the parent of the root directory_ doesn't mean anything beyond _go to the root directory_.

avarsava: Oh, didn't know a link to / would resolve, figured the link would be constructed like 'localhost:5000/single-lane-ts?run=my_cool_run/..'

PR #238, opened 2020-07-08
Simplest solution to provide proper ordering.

slazicoicr:
This tries to address the issue seen in #217, where more than one point having the same X-axis coordinate prevents true sorting.


avarsava (commented):
so I like the effect this has, but I think the extra number, especially displayed right beside the y value in the tooltip, is going to confuse users. Is having 2 rows with the same SampleName terribly common? Could we isolate just those ones and append a letter or even '_Run[n]' or something like that?

slazicoicr:
> 
> 
> so I like the effect this has, but I think the extra number, especially displayed right beside the y value in the tooltip, is going to confuse users. Is having 2 rows with the same SampleName terribly common? Could we isolate just those ones and append a letter or even '_Run[n]' or something like that?

Duplicated SampleName are the rule rather than the exception (most libraries are run on MiSeq and then on NovaSeq). Appending run info is going to make it extra long. Maybe:

* SampleName (1) or SampleName (ID1) or SampleName (PlotlyID1)
* Have the X-axis label be two lines using `<br>`. The second line then can be PlotlyID: 1 or the Run Alias

avarsava:
We will discuss data labeling in the meeting tomorrow, then I will take over this PR

avarsava:
@slazicoicr I've reworked quite a bit of this feature, in order to try to intelligently differentiate the x-axis from what we display as the x-axis. Please let me know if it needs more documentation, or if it doesn't make sense for some conditions (eg do we want this toggleable? If the First Sort is 'Sample' should it automatically toggle back?)

edit: Oh, and because the source branch isn't rebased on develop, the diff is terrible, sorry about that

application/dash_application/utility/plot_builder.py
```diff
@@ -264,8 +267,8 @@ def is_empty_plot(trace_list) -> bool:
 
 
 # writing a factory may be peak Java poisoning but it might help with all these parameters
-def generate(title_text, sorted_data, x_fn, y_fn, axis_text, colourby, shapeby,
-             hovertext_cols, cutoff_lines: List[Tuple[str, float]]=[],
+def generate(title_text, sorted_data, y_fn, axis_text, colourby, shapeby,
+             hovertext_cols, mode, x_fn=None, cutoff_lines: List[Tuple[str, float]]=[],
```
slazicoicr, line 271: A lot of code already depends on this change, but I would not change the signature of this function. Rather I would keep `x_fn` in it's place and allow it to be the `Mode` enum. This way we avoid having two parameters that compete with each other for what the x-axis should be. This would also allowed existing code from views to work. The code would be

```
if x_fn == Mode.IUS:
    # Same as your current IUS code
elif x_fn == Mode.MERGED:
    # Same as your current MERGED code
else:
    # Same as your current else code
````

File: application/dash_application/utility/plot_builder.py [hunk 2]
```diff
@@ -827,15 +846,29 @@ def __init__(
...
+        if self.x_col is None:
+            if self.mode == Mode.IUS:
+                self.x_fn = lambda d: d["SampleNameExtra"]
+                self.display_x = lambda d: d[PINERY_COL.SampleName]
+            elif mode == Mode.MERGED:
+                self.x_fn = lambda d: d[ml_col]
+                self.display_x = lambda d: d[ml_col]
+        else:
+            self.display_x = lambda d: x_fn(d)
+
```
slazicoicr, line 863: Would is make sense to move above code to `_generate_traces` (rather than `generate`), to avoid this code duplication?

application/dash_application/views/single_lane_rna.py
```diff
@@ -187,26 +188,24 @@ def get_rna_data():
      "value": PINERY_COL.SampleName}
 ]
 
-
+# TODO: Aww beans the subplots are gonna give me grief
```
slazicoicr: As all views will be switched over to subplots soon, the new ordering approach needs to work with subplots.

avarsava: that TODO was outdated, sorry for leaving it in :sweat_smile:

application/dash_application/utility/plot_builder.py
avarsava: Though Python allows it, I would rather not keep the signature the same. My reasoning is that aside from accepting two different data types on the same formal parameter being unorthodox, it does less to communicate to other developers that the results of setting the Mode are the default behaviours, and defining the x_fn is an override rather than an equally valid option. Were they on equal footing I would fear confusion over why setting x_fn to 'lambda d: d[PINERY_COL.SampleName]' isn't equivalent to setting it to Mode.IUS, and it would be less discoverable that setting it with a Mode is expected.
I have in this PR already adjusted every call to generate() to use this signature, no further refactoring is required.

application/dash_application/utility/plot_builder.py
avarsava: It is this way because the 'categoryarray' for the x-axis needs access to the correctly set x_fn.

slazicoicr: Ok, my main fear was that `generate` and here could become divergent. I think that `generate` won't be called anymore once we switch over to subplots, so this is fine.

application/dash_application/utility/plot_builder.py
```diff
@@ -827,15 +846,29 @@ def __init__(
...
+        self.mode = mode
 
     def traces(self):
+        self.display_x = None
+        self.x_fn = None
+        if self.x_col is None:
+            if self.mode == Mode.IUS:
+                self.x_fn = lambda d: d["SampleNameExtra"]
+                self.display_x = lambda d: d[PINERY_COL.SampleName]
+            elif mode == Mode.MERGED:
```
slazicoicr: Should be `self.mode`

application/dash_application/utility/plot_builder.py
```diff
@@ -827,15 +846,29 @@ def __init__(
...
+                self.x_fn = lambda d: d[ml_col]
+                self.display_x = lambda d: d[ml_col]
+        else:
+            self.display_x = lambda d: x_fn(d)
```
slazicoicr: Should be `self.x_fn`

application/dash_application/utility/plot_builder.py
```diff
@@ -559,16 +573,20 @@ def _define_graph(data, x_fn, y_fn, bar_positive, bar_negative, hovertext_cols,
     else:
         error_y = None
         hovertext_display_cols = hovertext_cols
-
     hovertext = create_data_label(data, hovertext_display_cols, additional_hovertext)
-
+    
+    additional_str = ""
+    if isinstance(hovertext, pandas.Series) and not hovertext.empty:
```
slazicoicr: I think any Python sequence will work here  (will silently throw away `hovertext` if user supplied it in Python `list` format), `if len(hovertext) > 0` should work.

slazicoicr (commented):
I think this will work (nice finding the `hovertemplate` solution). I think the cutoff lines hovertext needs to be fixed, as it shows the `SampleNameExtra` column data.

application/dash_application/utility/plot_builder.py
avarsava: Fixed

application/dash_application/utility/plot_builder.py
avarsava: Fixed

application/dash_application/utility/plot_builder.py
avarsava: Fixed

avarsava:
thanks Savo, could I get your opinions on
> if it doesn't make sense for some conditions (eg do we want this toggleable? If the First Sort is 'Sample' should it automatically toggle back?)

before I start rebasing onto the changes re: subplots? Let me know if you'd like to meet about it

slazicoicr:
For now, I would leave it as it is (all points always have their own x-position). Points sharing an x-position causes problems with the plots that contain error bars, as the error bars from two points perfectly overlap. I would not include a toggle feature yet. Can be implemented on request.

I still need to switch two call ready views onto subplots, but I am fine waiting for this to be merged into develop.

avarsava:
Hi Savo, was this closed by #258 ?

slazicoicr:
Yes.

PR #247, opened 2020-07-28
[GR-1221] Combine the two Call Ready TS views

slazicoicr:

PR #248, opened 2020-07-29
[GR-1222] Move single lange views to subplots

slazicoicr:


avarsava:
(Applies to GR-1221 as well) Axes start off showing -1 on the y-axis, which we previously completed a ticket for ensuring that doesn't happen.

avarsava (approved):

apmasell (approved):

slazicoicr:
Axes fix was done in #247 will apply to all subplots.

PR #252, opened 2020-08-11
[GR-1218] Display callability threshholds (and update to 0.38)

slazicoicr:


avarsava (commented):
Are there any points where the hovertext should say something other than 8/14? I can't find any

slazicoicr:
I checked. Currently, all data points are 8/14 in ETL. This was the default and the new workflow allows custom numbers. My guess is no project uses different setting at this point.

avarsava (approved):

apmasell (approved):

PR #265, opened 2020-09-21
[GR-1231] Thresholds

avarsava:
Many outstanding questions about these thresholds in the TODOs.
Clusters-Per-Sample commented out entirely due to no applicable column. 
Call-Ready WGS's Callability threshold is supposed to depend on whether 80x/30x coverage was used?
But works fine as it is, could be merged

### DIFF
--- a/application/dash_application/views/call_ready_rna.py
+++ b/application/dash_application/views/call_ready_rna.py
@@ -322,8 +322,8 @@ def layout(query_string):
                     # Cutoffs
                     sidebar_utils.cutoff_input(cutoff_insert_mean_label, ids["insert-mean-cutoff"],
                         initial["cutoff_insert_mean"]),
-                    sidebar_utils.cutoff_input(cutoff_clusters_per_sample_label, ids["clusters-per-sample-cutoff"],
-                        initial["cutoff_clusters_per_sample"]),
+                    # sidebar_utils.cutoff_input(cutoff_clusters_per_sample_label, ids["clusters-per-sample-cutoff"],
+                    #     initial["cutoff_clusters_per_sample"]),
                     sidebar_utils.cutoff_input(cutoff_rrna_contam_label, ids["rrna-contam-cutoff"],
                         initial["cutoff_rrna_contam"]),
                     sidebar_utils.cutoff_input(cutoff_percent_mapped_to_coding_label,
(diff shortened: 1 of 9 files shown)


apmasell (approved):
I think the code changes are good. If there's documentation about where these numbers come from any why (wiki page maybe?), can you reference it in a comment around where the cutoff are listed in the code?

slazicoicr (approved):

application/dash_application/views/call_ready_rna.py
```diff
@@ -412,20 +440,27 @@ def update_pressed(click,
...
             (cutoff_rrna_contam_label, special_cols["% rRNA Contamination"], rrna_contam_cutoff,
              (lambda row, col, cutoff: row[col] > cutoff)),
+            (cutoff_percent_mapped_to_coding_label, RNASEQQC2_COL.MetricsPercentCodingBases, percent_mapped_to_coding_cutoff,
+             (lambda row, col, cutoff: row[col] < cutoff)),
         ])
 
         new_search_sample = util.unique_set(df, PINERY_COL.RootSampleName)
 
         return [
+            #TODO: why this broken :(
```
slazicoicr: Old comment?

```suggestion
```

avarsava: Yes! Fixed

avarsava:
> I think the code changes are good. If there's documentation about where these numbers come from any why (wiki page maybe?), can you reference it in a comment around where the cutoff are listed in the code?

Talked with Morgan, got OK on linking google doc that specifies the thresholds. Pushed changes. Seem good?

apmasell:
Looks great!

application/dash_application/views/call_ready_wgs.py
```diff
@@ -282,7 +269,8 @@ def generate_deduplicated_coverage(df, graph_params):
         graph_params["colour_by"], 
         graph_params["shape_by"],
         graph_params["shownames_val"],
-        cutoff_lines=[(cutoff_coverage_tumour_label, graph_params[cutoff_coverage_tumour]), (cutoff_coverage_normal_label, graph_params[cutoff_coverage_normal])],
+        #TODO: Should these move to the CoverageMedian graph?
```
avarsava, line 272: @slazicoicr ?

slazicoicr: Median vs mean is still up in the air (also true of Insert Size Mean/Median). I will bring it up at tomorrow's meeting.

avarsava: to clarify - this is median vs 'deduplicated', same deal?

slazicoicr: Still same deal. Median coverage cannot be easily deduplicated from current BamQC data, so if cutoff needs to be on median coverage deduplicated (and not mean coverage deduplicated as it currently is), more work is required. Still want confirmation.

PR #268, opened 2020-09-22
[GR-1235] Use Clusters instead of Total Reads in single lane views

slazicoicr:


application/dash_application/views/single_lane_rna.py
```diff
@@ -71,13 +71,16 @@
 special_cols = {
     "Total Reads (Passed Filter)": "Total Reads PassedFilter",
```
avarsava, line 72: Is Total Reads still needed for anything?

avarsava (approved):
Changelog & wiki page update please!

application/dash_application/views/single_lane_rna.py
slazicoicr: Total Reads are still in the table.

apmasell (approved):

PR #269, opened 2020-09-23
[GR-1231] cfmedip thresholds

avarsava:
Most of them are commented out because we don't have enough information so effectively there's just one
Please see notes on Reporting Meeting 22/09/2020 on the wiki for explanation.

### DIFF
--- a/application/dash_application/views/single_lane_cfmedip.py
+++ b/application/dash_application/views/single_lane_cfmedip.py
@@ -78,8 +78,7 @@ initial = get_initial_cfmedip_values()
 initial["second_sort"] = CFMEDIP_COL.RelativeCpGFrequencyEnrichment
 initial["institutes"] = []
 # Set initial values for graph cutoff lines
-# TODO: is this just clusters-per-sample worded differently?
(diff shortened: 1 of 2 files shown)


slazicoicr (approved):

application/dash_application/views/single_lane_cfmedip.py
```diff
@@ -74,8 +78,17 @@
 initial["second_sort"] = CFMEDIP_COL.RelativeCpGFrequencyEnrichment
 initial["institutes"] = []
 # Set initial values for graph cutoff lines
-cutoff_pf_reads_label = sidebar_utils.total_reads_cutoff_label
-initial["cutoff_pf_reads"] = 0.01
+# TODO: is this just clusters-per-sample worded differently?
```
avarsava: @slazicoicr

apmasell (approved):

application/dash_application/views/single_lane_cfmedip.py
slazicoicr: Yes. Currently, the cluster column (from merging with FastQC) has not been done for this view.

PR #270, opened 2020-09-23
[GR-1231] Enable Clusters-per-sample minimums 

avarsava:
- The minimum requested from the lab is 10 000, but all the values in the graphs seem to be nowhere near that?
- Failures table on SL RNA specifically always says 'Missing', why?

### DIFF
--- a/application/dash_application/views/single_lane_rna.py
+++ b/application/dash_application/views/single_lane_rna.py
@@ -211,6 +211,7 @@ def generate_total_clusters(df, graph_params):
         graph_params["colour_by"],
         graph_params["shape_by"],
         graph_params["shownames_val"],
+        cutoff_lines=[{cutoff_clusters_per_sample_label, graph_params["cutoff_clusters_per_sample"]}]
     )
 
 def generate_insert_mean(df, graph_params):
@@ -437,8 +438,8 @@ def layout(query_string):
                     ids["insert-mean-cutoff"], initial["cutoff_insert_mean"]),
                 sidebar_utils.cutoff_input(cutoff_rrna_label,
                     ids["rrna-contamination-cutoff"], initial["cutoff_rrna"]),
-                # sidebar_utils.cutoff_input(cutoff_clusters_per_sample_label,
-                #     ids["clusters-per-sample-cutoff"], initial["cutoff_clusters_per_sample"]),
+                sidebar_utils.cutoff_input(cutoff_clusters_per_sample_label,
+                    ids["clusters-per-sample-cutoff"], initial["cutoff_clusters_per_sample"]),
 
                 html.Br(),
                 html.Button("Update", id=ids['update-button-bottom'], className="update-button"),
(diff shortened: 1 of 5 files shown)


apmasell (approved):

slazicoicr (approved):

avarsava:
> * The minimum requested from the lab is 10 000, but all the values in the graphs seem to be nowhere near that?
> 
> * Failures table on SL RNA specifically always says 'Missing', why?

@slazicoicr ?

slazicoicr:
10,000 clusters is suspiciously low. The lowest output we can do is ~1 million clusters, which means pooling 100 libraries, which happens very rarely. I will ask for clarification on Slack.

slazicoicr:
I am not seeing the "Missing" bug in the failure table when I pull up the DCRT project on Single Lane RNA-Seq.

@avarsava I think there is a bug with the Total Cluster cutoffs. The cutoffs is set to 10,000, but it is being compared to the column with is divided by 10^6. The cutoff needs to be either set to 0.01 or the value divided by 10^6 somewhere in Dashi before comparison.

avarsava:
> I am not seeing the "Missing" bug in the failure table when I pull up the DCRT project on Single Lane RNA-Seq.

'Missing' in several projects after pressing 'All Runs'.
 
> @avarsava I think there is a bug with the Total Cluster cutoffs. The cutoffs is set to 10,000, but it is being compared to the column with is divided by 10^6. The cutoff needs to be either set to 0.01 or the value divided by 10^6 somewhere in Dashi before comparison.

:+1: I will set it to 0.01. Are there any other metrics mismatched in this way?

application/dash_application/views/single_lane_rna.py
```diff
@@ -104,7 +104,8 @@
 cutoff_insert_mean_label = sidebar_utils.insert_mean_cutoff_label
 initial["cutoff_insert_mean"] = 150
 cutoff_clusters_per_sample_label = sidebar_utils.clusters_per_sample_cutoff_label
-initial["cutoff_clusters_per_sample"] = 10000
+# This is 10 000, but the stat is / 10^6
+initial["cutoff_clusters_per_sample"] = 0.01
```
slazicoicr, line 108: Change the cutoff label to reflect that the entered value will be 10^6 clusters.

slazicoicr:
> 👍 I will set it to 0.01. Are there any other metrics mismatched in this way?

No.

application/dash_application/views/single_lane_rna.py
avarsava: Done

slazicoicr:
Ok, I see the `Missing` values. I picked `PCSI_0692_Pa_P_PE_311_WT` to investigate.

`Missing` means that the value is empty (`pandas.nan`). This happened because there is no FastQC file for that library (which is used to calculate clusters). I get no hits when I run ` zcat seqware_files_report_latest.tsv.gz | grep PCSI_0692_Pa_P_PE_311_WT | grep -i fastqc`.

slazicoicr (approved):

apmasell (approved):

PR #273, opened 2020-10-01
[GR-1230] Send request to MISO

avarsava:
This is a draft because I want feedback before adding this feature to all the other reports. This is currently only implemented on Single-Lane TS. Looking for assistance in particular on the way the CSS causes the buttons to be slightly misaligned on the right side of the sidebar, and whether anything needs more documentation.


application/dash_application/assets/styles.css
```diff
@@ -377,6 +373,13 @@ input[type="button"] {
     box-sizing: border-box;
 }
 
+.sidebar .miso-qc-button {
+    float: left;
+    width: 30%;
+    margin-left: 2.5%;
+    margin-right: 2.5%;
```
avarsava, line 380: I know 3(30) + 2(2.5) != 100, but when I set these to 5, they'd overflow, and i couldn't identify the margin or padding that was ruining it

apmasell: CSS is supposed to make sense? If it bothers you, maybe `calc`. I'm good with it.

apmasell (approved):

application/dash_application/utility/df_manipulation.py
```diff
@@ -475,3 +475,37 @@ def remove_suffixed_columns(df: DataFrame, suffix: str) -> DataFrame:
...
+    Expects an array of dicts in the format:
+    [{
+        'title': threshold name minus the 'maximum' or 'minimum', 
+        'threshold_type': gt, ge, lt, le
+        'threshold': the current threshold setting from the sidebar,
+        'value': the column to check in the dataframe for this threshold
+    }]
+    Returns [json string dump of http body, button style]
+    """
+    if (len(df.index) > 0):
```
slazicoicr, line 491: This is equivalent to `len(df) > 0`

application/dash_application/utility/df_manipulation.py
```diff
@@ -475,3 +475,37 @@ def remove_suffixed_columns(df: DataFrame, suffix: str) -> DataFrame:
...
+                metrics_this_row.append({
+                    'title': metric['title'],
+                    'threshold_type': metric['threshold_type'],
+                    'threshold': metric['threshold'],
+                    'value': row[metric['value']]
+                })
+            miso_request['library_aliquots'].append({
+                'name': split_provenance_id[2],
+                'metrics': metrics_this_row,
+                'run': split_provenance_id[0]
```
slazicoicr: Would this require `lane` as well?

application/dash_application/views/single_lane_ts.py
```diff
@@ -524,6 +527,15 @@ def update_pressed(click,
 
         (jira_href, jira_style) = sidebar_utils.jira_display_button(runs, title)
 
+        (miso_request, miso_button_style) = util.build_miso_info(df, title, 
+            [{
+                'title': 'Passed Filter Reads (*10^6)',
```
slazicoicr: Should this be `Total Clusters`?

application/dash_application/utility/df_manipulation.py
avarsava: Shouldn't, this is the format @djcooke specified for receiving into MISO.
Dillan, should I be sending MISO the lane as well?

application/dash_application/utility/df_manipulation.py
avarsava: it is, but len(df.index) is slightly more performant: https://stackoverflow.com/a/18317067

application/dash_application/views/single_lane_ts.py
avarsava: Looks like part of SLTS got overlooked when making this switch. I'll fix it in another PR then rebase this and fix

application/dash_application/utility/df_manipulation.py
djcooke: Depends how specific you want it to be:

* If you provide only the library aliquot, MISO will include all lanes that it has been run in
* If you provide run + library aliquot, MISO will filter to include only lanes from that run that include the library aliquot
* If you provide run + lane + library aliquot, MISO could filter to include only that lane

If the metrics are specific to a lane, then you probably want to provide the lane.

avarsava: @slazicoicr Metrics are specific to a Lane because of how IUSes work, right?

avarsava: Updated

avarsava:
Demonstrated to work with the MISO half! time to expand to the other reports and open for review

apmasell (approved):

slazicoicr (approved):

PR #274, opened 2020-10-02
[GR-1235] Call ready views use clusters instead of reads

slazicoicr:
I will update Wiki.

Also fixed up Single Lane TS to remove Total Read vestiges as pointed out in https://github.com/oicr-gsi/dashi/pull/273#discussion_r498838582

### DIFF
--- a/application/dash_application/utility/sidebar_utils.py
+++ b/application/dash_application/utility/sidebar_utils.py
@@ -12,7 +12,6 @@ import pinery.column
 import gsiqcetl.column
 from . import df_manipulation as df_tools
 
-total_clusters_cutoff_label = "Passed Filter Clusters (* 10^6) minimum"
 insert_mean_cutoff_label = "Mean Insert Size minimum"
 insert_median_cutoff_label = "Median Insert Size minimum"
 percent_duplication_cutoff_label = "% Duplication maximum"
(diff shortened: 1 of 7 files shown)


application/dash_application/utility/sidebar_utils.py
```diff
@@ -12,7 +12,7 @@
 import gsiqcetl.column
 from . import df_manipulation as df_tools
 
-total_reads_cutoff_label = "Passed Filter Reads (* 10^6) minimum"
+total_clusters_cutoff_label = "Passed Filter Clusters (* 10^6) minimum"
```
avarsava: Is this different from clusters_per_sample_cutoff_label at line 19? 'Clusters Per Sample' was the wording requested in the requirements document.

application/dash_application/views/call_ready_rna.py
```diff
@@ -66,7 +66,8 @@ def dataversion():
     "Total Reads (Passed Filter)": "Total reads passed filter",
```
avarsava, line 66: Is the Total Reads special_col still required?

application/dash_application/utility/sidebar_utils.py
slazicoicr: Fixed.

application/dash_application/views/call_ready_rna.py
slazicoicr: This way both reads and cluster metrics will be in the table.

avarsava (approved):

apmasell (approved):

PR #278, opened 2020-10-08
set miso url to stage

avarsava:
Will reset this to prod once feature is released to prod


apmasell (approved):

slazicoicr (approved):
Docker is still somewhat of a mystery to me, but there seem to be two ways that environmental variables are being set: https://bitbucket.oicr.on.ca/projects/GSI/repos/spb-seqware-production/browse/dashi/docker-compose.yml#48

avarsava:
> Docker is still somewhat of a mystery to me, but there seem to be two ways that environmental variables are being set: https://bitbucket.oicr.on.ca/projects/GSI/repos/spb-seqware-production/browse/dashi/docker-compose.yml#48

Huh, you're right. Any opinions on where this particular addition should go?

PR #286, opened 2020-10-23
[GR-1237 GR-1241] Median calculation fixes

slazicoicr:


apmasell (approved):

avarsava:
The median insert size graph has a greater number of strange bars in it with this change compared to qc-stage. Should that change with a data update?

slazicoicr:
That is unexpected. Which view/project/run are you seeing this with? Did you run `pip install -r requirements.txt` to update ETL version first?

avarsava:
Currently running gsiqcetl 0.43.2 against production data. The run 200923_NB551056_0158_AHMG2HBGXF has 6 strange bars with this change but only 1 on qc-stage.
image
(Another run was added just for comparison)

slazicoicr:
Those are valid representations of the Picard InsertMetrics data (https://jira.oicr.on.ca/browse/GC-8187). The previous version bug actually hid the crazy values. They are due to bad sequencing.

avarsava (approved):

PR #299, opened 2020-12-10
Add removed columns back

avarsava:
I removed these columns from the raw data tables 5 months ago when we decided Insert Size should go. Now it's back and should be in the raw data table again.


slazicoicr:
Initiated wider discussion at https://oicr.slack.com/archives/C018V6EQG4U/p1607620479007000

asofmosq (approved):

avarsava:
@slazicoicr changed names

slazicoicr (approved):

PR #300, opened 2020-12-14
[GR-1264] Fixed incorrect Coverage in Call Ready WGS Failed Sample table

slazicoicr:


avarsava (approved):

avarsava:
Does 'None' deliver correctly to the MISO Run-Library QC page?

slazicoicr:
The `None` feature is is Call Ready pages, which (I believe) do not have a link to MISO QC page.

avarsava (approved):
Might necessitate correcting the handling of None on the MISO side when I get to implementing QC in MISO for Call Ready reports but that can wait

slazicoicr:
In this case, `None` mean Not Applicable. The QC metric in this column does not apply to the sample in this row.

asofmosq (approved):

PR #303, opened 2021-01-11
Status page that included ETL info (date modified/errors)

slazicoicr:
Accessed at 127.0.0.1:5000/status. Where should I put the link to this? In the footer with the Dashi version?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and as of version 1.0.0, follows semantic versioning.
 
 ## [Unreleased]
+## Added
+  * `/status` page showing date of the latest cache and cache errors
 
 ## [210104-1505] - 2021-01-04
 ## Changed
--- a/application/routes.py
+++ b/application/routes.py
@@ -77,7 +77,10 @@ def status_page():
         if os.path.exists(error_path):
             with open(error_path, "r") as f:
                 j = json.load(f)
-                errors[cache.name] = json.dumps(j, indent=2)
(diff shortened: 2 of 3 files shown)


apmasell (approved):

avarsava (approved):
I'm not sure how to interpret the page. What is the difference between 'no error file' and '[]'? 
With regards to where to link this page: if there were entries in that array what would they be telling me? Would regular users find said information useful? Should this move to an endpoint for the gsi-qc-etl web service once that's done?

slazicoicr:
`No error file` means the error file does not exist on the disk. This is because the cache has not been updated since that feature was implemented in ETL. `[]` means there are currently no errors (empty list). I can change `[]` to `No Errors`. If there were errors, it would be the failed Shesmu data send to ETL.

This is definitely a gsi-qc-etl web service feature that's piggybacking on the Dashi server. The use would be troubleshooting missing records. @avarsava let me know if I should make any changes to this or go ahead as is.

avarsava:
> `No error file` means the error file does not exist on the disk. This is because the cache has not been updated since that feature was implemented in ETL. `[]` means there are currently no errors (empty list). I can change `[]` to `No Errors`. If there were errors, it would be the failed Shesmu data send to ETL.
> 
> This is definitely a gsi-qc-etl web service feature that's piggybacking on the Dashi server. The use would be troubleshooting missing records. @avarsava let me know if I should make any changes to this or go ahead as is.

'No Errors' or even 'OK' would be clearer, I think. I don't know whether users would know what to do with the information that Shesmu failed to send data to ETL, so I'm OK with the url being folklore and not linked anywhere in Dashi

PR #304, opened 2021-01-15
[GR-1272] Alert message when requested runs do not exist

slazicoicr:


apmasell:
Nitpick: "No data for run" rather than "Doesn't exist".

apmasell (approved):

avarsava (commented):
- Please add warning to SARS-CoV-2 report as well (it's not labeled as such but it's a single-lane report too)
- Code repetition could be reduced with a function taking initial["runs"], ALL_RUNS, and returning the Alert

avarsava:
image
Could there be a `<br>` after the alert so it doesn't push over the JIRA buttons?

slazicoicr:
Hmm, `<br>` does not seem to work (due to `row jira-buttons` div having flex display). How about putting the alert after the two boxes?

image

avarsava:
> Hmm, `<br>` does not seem to work (due to `row jira-buttons` div having flex display). How about putting the alert after the two boxes?
> 
> image

I like it!

avarsava (approved):

PR #313, opened 2021-04-09
[GR-1281] Gracefully dealing with missing caches

slazicoicr:
image

### DIFF
--- a/application/dash_application/pages.py
+++ b/application/dash_application/pages.py
@@ -4,6 +4,7 @@ import sys
 import traceback
 
 import dash_html_components as html
+from .utility import sidebar_utils
 
 prefix = "application.dash_application.views."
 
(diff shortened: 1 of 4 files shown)


application/dash_application/pages.py
```diff
@@ -29,6 +34,17 @@
 
 # Please do not edit this loop
 for name in pagenames:
-    pages.append(importlib.import_module(prefix + name))
+    try:
+        pages.append(importlib.import_module(prefix + name))
+    except (IOError, OSError):
+        print(traceback.format_exc(), file=sys.stderr)
+        ErrorPage = namedtuple('ErrorPage', 'layout title dataversion page_name init_callbacks')
```
apmasell: This should probably go somewhere more top-level.

apmasell (approved):

application/dash_application/pages.py
```diff
@@ -29,6 +34,17 @@
...
+        pages.append(ErrorPage(
+            lambda x: html.H3("Unexpected Error: Failed to load cache"),
```
callunity: When this bug appears in the wild, it'll take a developer to fix it. Can we either add Prometheus monitoring that fires an alert for this case, or add a button for the user to submit a ticket describing the issue?

application/dash_application/utility/df_manipulation.py
```diff
@@ -136,32 +136,9 @@ def normalized_merged(df: DataFrame, merged_cols: List[str]):
 
 
 """
-Open a single instance of each cache, and use copies for the reports.
+qc-etl API keeps loaded caches in memory for fast reloading
```
callunity, line 139: Out of curiosity, how much memory does this take at present?

slazicoicr: I am not sure. `flask` takes 1.2GB on my machine.

callunity (approved):

application/dash_application/pages.py
avarsava: Echoing this

application/dash_application/pages.py
slazicoicr: Done

application/dash_application/pages.py
slazicoicr: Error view contains a link that will generate a JIRA ticket.

PR #325, opened 2021-07-23
[GR-1295] Runscanner Illumina Flow Cell view for Q30

slazicoicr:
This requires the development cache of QCETL as runscanner is not enabled in prod yet. This view is minimal. The idea is to have a view working on dev to get user feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and as of version 1.0.0, follows semantic versioning.
 
 ## [Unreleased]
+## Changed
+  * `fill_in_color_col` and `fill_in_shape_col` take an arbitrary column for color/shape
+
+## Added
+  * Added Runscanner Illumina Flow Cell view
 
 ## [210531-1427] - 2021-05-31
   * Remove bcl2barcode WIP and fix x-axis being cut off/missing tick labels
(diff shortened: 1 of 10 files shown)


avarsava:
The official branding is 'Run Scanner'

avarsava:
I'm not sure 'bool or string' is a type that makes sense for a parameter called 'call_ready'. In this design 'call_ready' = 'RunScannerCol.RUN'.

Could we instead have a globally accessible dictionary of 'report types' where the type of report is defined at the top of each report (eg 
```
import ReportTypes

TYPE = ReportTypes["Call-Ready"]
```
and then further down
`return plotbuilder.Subplot([.......], TYPE)`

Where ReportTypes might be
```
ReportTypes = {"RunScanner": RUN_COL.Run, 
"Single-Lane": PINERY_COL.SampleName,
"Call-Ready": PINERY_COL.RootSampleName}
```
and then rename the 'call_ready' param to something more clear, like 'xaxis_col'?

But this also highlights how unlike all the other reports this one is. Do any of the other views report on hardware metrics rather than sample qc data? Since it's not like one can interact with sample qc data at the same time on this report, I'm wondering if Dashi is a place that makes sense for this.

slazicoicr:
I agree regarding bool/str (gotta love Python) and can make the function just use str. 

As to if this belongs in Dashi, I want to see what users want from this view. If it is QC only and has the same properties as the other views (consistent X-axis, color, shape, filtering, cut-off points) then it makes sense to have machine QC in the same place. Machine QC are part of our QC pipeline and if a given run in Single Lane views looks really strange, looking at machine QC is part of troubleshooting.

avarsava (approved):

callunity (approved):

PR #326, opened 2021-08-27
[GR-1295] Turn of runscanner view until refiller bug is fixed

slazicoicr:
Disable for now so new features can be released without broken runscanner page.


avarsava (approved):
In the future, let's keep unfinished features on their own branches and deploy those branches to dev when we need them

slazicoicr:
Yeah, this one was strange. It works on Dashi dev because Shesmu dev refiller works. I should have taken the refiller through deployment first (would have noticed the Shesmu bug), but I did Dashi and refiller deployment in parallel which was the mistake.

callunity (approved):

PR #328, opened 2021-09-09
Use env var to allow getting all Pinery projects

callunity:
This keeps "active projects only" as the default, but allows it to be overriden by an environment variable

### DIFF
--- a/application/dash_application/utility/df_manipulation.py
+++ b/application/dash_application/utility/df_manipulation.py
@@ -1,3 +1,4 @@
+import os
 import pandas
 from pandas import DataFrame, Series
 from typing import List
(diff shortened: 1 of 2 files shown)


avarsava (approved):

application/dash_application/utility/df_manipulation.py
```diff
@@ -370,11 +371,12 @@ def get_rnaseqqc2_merged():
 def get_pinery_samples(active_projects_only=True):
```
slazicoicr, line 371: The optional `active_projects_only` is never overridden. I think it's cleaner to just remove this parameter and have the environmental variable be the decider. However, I can totally see a future where one Dashi view needs to have all projects and all the other ones don't.

slazicoicr (approved):

application/dash_application/utility/df_manipulation.py
callunity: Yes, that very possible future was why I left it in as well :grimacing:

PR #330, opened 2021-09-29
[GC-8958] Swap workflow has been paused.

slazicoicr:
As the workflow is paused, view won't be updated. This informs the user of that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and as of version 1.0.0, follows semantic versioning.
 
 ## [Unreleased]
+## Changed
+  * Added status to swap view that workflow is paused.
 
 ## [210914-1010] - 2021-09-14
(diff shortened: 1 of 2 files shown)


avarsava:
How long is that ticket expected to be open? If it's going to be quick I'm not sure embedding this message in the code base is the best course, since this will require a release to roll out and another release to undo

slazicoicr:
@mlaszloffy Can you comment on Alexis's concern?

mlaszloffy:
We'll have a better idea on how long this will take after friday. Hold off until monday?

slazicoicr:
@mlaszloffy From our brief chat on Friday, I think a quick fix is not possible for this workflow. Does it make sense to go ahead with this PR?

slazicoicr:
There is confusion about the Dashi view not being up to date: https://jira.oicr.on.ca/browse/GRDQ-113

slazicoicr:
Everything is up again. A message display feature will be implemented, so that user facing messages do not require Dashi release.

PR #336, opened 2021-11-11
[GR-1320] Display user messages on individual views.

slazicoicr:
It was not possible to inform users of issues with Dashi without a Dashi
release. This solves the problem by allowing a local JSON file to hold
messages displayed on individual views.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Create a `.env` file in the root directory of this repository:
 | `USE_BLEEDING_EDGE_ETL` | No | Set to install `gsi-qc-etl@master` instead of the release version of `gsi-qc-etl` in `requirements.txt` (Docker only) | `1` | use release version |
 | `EXCLUDE_SWAP_LIBS` | No | File path to TSV file of library pairs to be excluded for swap view | `./exclude_swap_lib.tsv` | |
 | `SAMPLES_FOR_PROJECTS` | No | Indicate whether samples from ALL projects should be used, or only samples from ACTIVE projects. | `ALL` | `ACTIVE` |
+| `DISPLAY_USER_MESSAGE` | No | A JSON file containing a dictionary of page names (key) and messages to display (value)| `./user_messages.json` | |
 
 ## Setup on bare metal
 
--- a/application/dash_application/known_pages_router.py
+++ b/application/dash_application/known_pages_router.py
@@ -68,6 +68,13 @@ def navbar(current):
 
 
 def load_user_messages(json_path=os.getenv("DISPLAY_USER_MESSAGE")):
+    """
+    Loads the user messages. Raises an exception in the following cases:
+    * Cannot parse JSON file
+    * If message is not a string
+    * If the view of the message does not exist
+    * If multiple messages exist for a single view
+    """
     if json_path is None:
         return {}
 
@@ -75,9 +82,25 @@ def load_user_messages(json_path=os.getenv("DISPLAY_USER_MESSAGE")):
         logger.warning("User message JSON file does not exist")
         return {}
 
-    # If opening or decoding fails, Dashi will crash.
+    def json_parse(pairs):
+        parsed = {}
+
+        for k, v in pairs:
+            if not isinstance(v, str):
+                raise TypeError("User messages must be a string: {}".format(v))
+
+            if k not in pages_info:
+                raise KeyError("Unknown page name: {}".format(k))
+
+            if k in parsed:
+                raise KeyError("Multiple messages for the same page: {}".format(k))
+            else:
+                parsed[k] = v
+
+        return parsed
+
     with open(json_path, "r") as f:
-        return json.load(f)
+        return json.load(f, object_pairs_hook=json_parse)
 
 
 # Messages to the displayed to users in specific views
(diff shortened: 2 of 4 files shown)


application/dash_application/known_pages_router.py
```diff
@@ -113,3 +148,22 @@ def content_handler(path, qs):
...
+            Output('user_message', 'is_open'),
+        ],
+        [
+            Input('url', 'pathname'),
+        ])
+    def display_user_message(path):
+        """
+        If there is a user message associated with the loaded page, display it.
+        """
+        requested = path[1:] # drop the leading slash
```
avarsava, line 184: I'm having trouble with this line. It stack traces here with `TypeError: 'NoneType' object is not subscriptable` on every page load, even when my json file is well formed and contains a banner message that displays correctly.

README.md
```diff
@@ -30,6 +30,7 @@ Create a `.env` file in the root directory of this repository:
 | `USE_BLEEDING_EDGE_ETL` | No | Set to install `gsi-qc-etl@master` instead of the release version of `gsi-qc-etl` in `requirements.txt` (Docker only) | `1` | use release version |
 | `EXCLUDE_SWAP_LIBS` | No | File path to TSV file of library pairs to be excluded for swap view | `./exclude_swap_lib.tsv` | |
 | `SAMPLES_FOR_PROJECTS` | No | Indicate whether samples from ALL projects should be used, or only samples from ACTIVE projects. | `ALL` | `ACTIVE` |
+| `DISPLAY_USER_MESSAGE` | No | A JSON file containing a dictionary of page names (key) and messages to display (value)| `./user_messages.json` | |
```
avarsava, line 33: Could we include a user_messages.json-example like we do for .env?

application/dash_application/known_pages_router.py
```diff
@@ -63,10 +67,41 @@ def menu_item(label, link):
...
 
+def load_user_messages(json_path=os.getenv("DISPLAY_USER_MESSAGE")):
+    if json_path is None:
+        return {}
+
+    if not os.path.isfile(json_path):
+        logger.warning("User message JSON file does not exist")
+        return {}
+
+    # If this fails, it will take down Dashi, so general exception handling is a must
```
avarsava: Should a problem with the message file maybe take down Dashi? Unless we add alerting when these warnings are found in Loki, problems with the message file will fail relatively silently and we might not check that the banner is showing up correctly in prod. If the service crashes then at least we'll hear about it

avarsava (commented):
- I would prefer a higher contrast font colour, the brown on yellow seems non-urgent to me
- I would like if we could match the padding and margins of the surrounding elements: 
image

avarsava:
It's possible to make a blank banner by setting eg "single-lane-ts":true

avarsava:
if the same report is used as a key twice, one message is picked for display. Testing is too slow to tell whether it's random or always the bottom one

slazicoicr:
> 
> 
> if the same report is used as a key twice, one message is picked for display. Testing is too slow to tell whether it's random or always the bottom one

From the specs: `The RFC specifies that the names within a JSON object should be unique, but does not mandate how repeated names in JSON objects should be handled. By default, this module does not raise an exception; instead, it ignores all but the last name-value pair for a given name`. Should an exception be raised?

avarsava:
"single-lane-ts":["test1", "test2"]
renders on the page as 
image
I don't know why anyone would do most of these things to be fair

avarsava:
> > if the same report is used as a key twice, one message is picked for display. Testing is too slow to tell whether it's random or always the bottom one
> 
> From the specs: `The RFC specifies that the names within a JSON object should be unique, but does not mandate how repeated names in JSON objects should be handled. By default, this module does not raise an exception; instead, it ignores all but the last name-value pair for a given name`. Should an exception be raised?

If our list ends up growing to a length that's hard to hold entirely in mental memory, throwing an exception in this case might be useful. I think we should avoid having messages getting ignored silently

avarsava:
adding `"secret": "you found the secret page"` actually displays the message on the 404 page for <dashi url>/secret ! I don't know if we should check that the page named actually exists or not. this behaviour might cause entries with typo'd keys to fail silently.

application/dash_application/known_pages_router.py
slazicoicr: I changed the message to be red.

asofmosq (approved):

application/dash_application/known_pages_router.py
```diff
@@ -63,10 +67,58 @@ def menu_item(label, link):
...
+
 # Default layout element (wraps the page layout elements which are returned by the router)
 layout = html.Div([
     core.Location(id='url', refresh=False),
     navbar(default_title),
+    dbc.Alert(
+        id="user_message",
+        is_open=False,
+        color="danger",
+        style={"margin-left": "15px"},
```
avarsava: margin-right too please

user_messages.json
```diff
@@ -0,0 +1,3 @@
+{
+    "single-lane-rna": "Report is amazing. Just wanted to let you know."
```
avarsava, line 2: :grin:

avarsava (approved):

PR #344, opened 2022-01-31
[GR-1344] Load sample provenance from disk rather than Mongo DB

slazicoicr:
Couldn't put `MONGO_FILE` in the docker configs on bitbucket. it would conflict with MONGO_URL in `start.sh`. Can't delete MONGO_URL ahead of time, as a Mongo source is mandatory.

### DIFF
--- a/.docker/start.sh
+++ b/.docker/start.sh
@@ -16,6 +16,6 @@ fi
 # mount your private key for bitbucket into the container to allow this command to succeed
 pip install --trusted-host pypi.python.org -r requirements.txt
 
-export MONGO_FILE=/scratch2/groups/gsi/production/qcetl/provenance_samples/latest
+export MONGO_FILE=/cache/provenance_samples/latest
 
 flask run --host=0.0.0.0 --port=5000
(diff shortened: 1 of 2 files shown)


.docker/start.sh
```diff
@@ -16,6 +16,6 @@ fi
 # mount your private key for bitbucket into the container to allow this command to succeed
 pip install --trusted-host pypi.python.org -r requirements.txt
 
-export MONGO_URL="mongodb://provenance_ro:$(cat /run/secrets/mongo_password)@provenance-mongo-db.gsi.oicr.on.ca:27017/provenance"
+export MONGO_FILE=/scratch2/groups/gsi/production/qcetl/provenance_samples/latest
```
mlaszloffy, line 19: I think this should be `/cache/provenance_samples/latest`?

slazicoicr:
Sorry, I deleted @mlaszloffy comment. (about the wrong "MONGO_FILE" path. I think it should be: `/home/ubuntu/qcetl/provenance_samples`

avarsava:
What effect does this have on GR-1140?

slazicoicr:
Re GR-1140:

- This is merged (need to remove MONGO_URL)
- Add MONGO_FILE to bitbucket (won't conflict)
- Remove MONGO_FILE from start.sh

avarsava (approved):

mlaszloffy (approved):

PR #345, opened 2022-02-04
[GR-1349] Unticking "only show swaps" shows all projects.

slazicoicr:
Unticking "only show swaps" only showed projects that had swaps detected. This was
counter-intuative. Unticking now shows all projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and as of version 1.0.0, follows semantic versioning.
 ## [Unreleased]
 ## Changed
   * Sample provenance is loaded from cache on disk rather than Mongo DB
+  * Unticking `only show swaps` shows all projects in swap view
+  * Added project name to swap view (does not always match Alias)
+  * Showing meta data along library alias in swap view
 
 ## [220125-1411] - 2022-01-25
 ## Changed
--- a/application/dash_application/views/sample_swaps.py
+++ b/application/dash_application/views/sample_swaps.py
@@ -232,10 +232,7 @@ for d in TABLE_COLUMNS:
 
 
 # Pair-wise comparison is done within project (for now), so left project is sufficient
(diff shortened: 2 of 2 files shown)


avarsava:
I've deployed this change locally and I can't identify the change in behaviour compared to qc-dev, and I'm not sure I understand your description. Unticking the box doesn't seem to affect the project filter widget?

slazicoicr:
One you untick "only show swaps" and click Update, the project filter drop-down does not change what is selected, but should now contain all projects for a user to select.

The other option was for unticking to automatically select all projects, but that would cause strange behaviour if a user selected only one project (wants to see all samples within that one project) and unticked the box..

mlaszloffy (approved):

avarsava:
I thought projects only get shown on this report if there are swaps?

slazicoicr:
By default yes. But it is necessary to see metrics of samples not marked as swaps (both to investigate false negatives and to reason about complex swaps).

This PR came about because Alex wanted to investigate a false negative (he thought it was a swap, but it did not show up in the Dashi view). Unticking "only show swaps" would show you all data for a project where there is at least one swap. However, there was no way to investigate project where there were no swaps. This fixes that.

avarsava:
I see. Do any other reports operate in this way, where the total set available in the projects dropdown changes based on other widgets? I'd almost want to always show all the projects as options in the dropdown, and untie the check box from the projects filter. And if you've selected a project with no swaps and check 'only show swaps', you get an empty report, which should be good news for you

slazicoicr:
That makes a lot of sense. All projects shown by default now.

avarsava (approved):
Looks good. We'll just want to make sure users understand the change because now when the page loads there's 2x as many projects preselected in the widget, which might look like an explosion of swaps if they don't understand

slazicoicr:
Sounds good. I'll post on the swap Slack channel.

PR #389, opened 2023-04-05
Fixed dependency versions to speed up `pip install`

slazicoicr:
Jira ticket or GitHub issue:

- [X] Updates CHANGELOG.md
- [X] Updates dev docs if applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and as of version 1.0.0, follows semantic versioning.
 
 ## [Unreleased]
+  * Fixed dependency versions to speed up `pip install`
 
 ## [230330-1838] - 2023-03-30
   * Reverted to rnaseqqc2 V2 cache due to column naming bug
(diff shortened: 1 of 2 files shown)


avarsava:
Could these be `>=` instead of `==`? Might miss bugfixes and patches this way

slazicoicr:
I tried restricting it to a major version, while allowing minor and patch updates and I still get a lot of `pip is looking at multiple versions`. I think each package has lots of dependencies and we either freeze them exactly or `pip` has to resolve them.

But the actual install took less than 1 minute on my computer even with resolving the dependencies, so I am fine with allowing for minor/patch updates.

Restricting to major version is done by `dash~=2.8` as per https://peps.python.org/pep-0440/#compatible-release

avarsava (approved):

callunity:
> I tried restricting it to a major version, while allowing minor and patch updates and I still get a lot of `pip is looking at multiple versions`. I think each package has lots of dependencies and we either freeze them exactly or `pip` has to resolve them.
> 
> But the actual install took less than 1 minute on my computer even with resolving the dependencies, so I am fine with allowing for minor/patch updates.
> 
> Restricting to major version is done by `dash~=2.8` as per https://peps.python.org/pep-0440/#compatible-release

I agree that restricting to major version makes sense. It's fine if pip takes some time to do the dependency resolution; I was hoping to reduce the time, rather than eliminating it entirely.

callunity (approved):

PR #390, opened 2023-04-05
Allow Dashi to pull from more than one cache source

slazicoicr:
The `/status` end point has been designed with only one source in mind. It requires a redesign to deal with multiple caches. As the `/status` endpoint is used for troubleshooting and archival caches only contain finished cases, this is a future todo  GR-1496 .

Jira ticket or GitHub issue:

- [x] Updates CHANGELOG.md
- [x] Updates dev docs if applicable


avarsava:
I don't understand. I thought the design was that qc-etl pulls from multiple sources including an archive and presents all the aggregated data as one, so Dashi would not need to know the difference

slazicoicr:
That's true for the metrics collected in the caches. The metadata about those metrics (when they were collected, what the Shesmu input was, which ones failed) hasn't received as much love. Fixing that will be gsiqcetl V2 territory.

It's that metadata that `/status` endpoint presents.

README.md
```diff
@@ -20,7 +20,7 @@ Create a `.env` file in the root directory of this repository:
 
 | Variable name               | Required?              | Description                                                                                                           | Example                                               | Default |
 |-----------------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------|---------|
-| `GSI_QC_ETL_ROOT_DIRECTORY` | **Yes**                | Directory where the QC-ETL caches are located                                                                         | `/qcetl`                                              | |
+| `GSI_QC_ETL_ROOT_DIRECTORY` | **Yes**                | One or more colon seperated directories where the QC-ETL caches are located                                           | `/qcetl` or `/qcetl:/qcetl_archive`                    | |
```
callunity: For the archival db-backed cache, is this the directory where the metadata is located?

callunity: I think it would be valuable to specify here that in case of duplicate records, the records from the directory/cache listed first will be selected.

callunity:
> archival caches only contain finished projects

(just as a small note that doesn't affect any of the contents of this PR: this won't always be true - we'll be adding to the archival caches as cases from active projects complete)

README.md
slazicoicr: > For the archival db-backed cache, is this the directory where the metadata is located?

Correct. The only requirement for gsiqcetl is the `config.json`, which specifies how to connect to the database.

callunity (approved):

avarsava (approved):

PR #425, opened 2024-09-30
Fix miso url

avarsava:
Jira ticket or GitHub issue:

- [X] Updates CHANGELOG.md
- [ ] Updates dev docs if applicable

### DIFF
--- a/application/dash_application/utility/sidebar_utils.py
+++ b/application/dash_application/utility/sidebar_utils.py
@@ -432,7 +432,7 @@ def miso_qc_button(body_id, button_id):
         core.Input(id=button_id, value="QC in MISO", type="submit", className="miso-qc-button")
     ],
     method="POST",
-    action=os.getenv("MISO_URL")+"miso/runlibraries/metrics",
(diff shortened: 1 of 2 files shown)


application/dash_application/utility/sidebar_utils.py
```diff
@@ -432,7 +432,7 @@ def miso_qc_button(body_id, button_id):
         core.Input(id=button_id, value="QC in MISO", type="submit", className="miso-qc-button")
     ],
     method="POST",
-    action=os.getenv("MISO_URL")+"miso/runlibraries/metrics",
+    action=os.getenv("MISO_URL")+"runlibraries/metrics",
```
djcooke, line 435: Does this assume there's a front-slash ('/') at the end of `MISO_URL`? If so, the front-slash should be removed in the other two places that now use `MISO_URL`

avarsava: fixed

djcooke (approved):

wuall826 (approved):

PR #450, opened 2026-08-14
Gp 5767 poc call ready wgs ica report

riyapatel36:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GP-5767

- [x] Updates CHANGELOG.md
- [ ] Updates dev docs if applicable


avarsava:
LGTM as a template for handing off to UHN, but I don't think this needs to be checked into develop.

riyapatel36:
> LGTM as a template for handing off to UHN, but I don't think this needs to be checked into develop.

Even if this merges into develop, it won't affect our own Dashi deployment. I've already added an ENABLED_REPORTS env var that controls which report pages actually load — if it's unset, all reports run as before, so our deployment is unaffected.

For our setup we'd set it to include everything:

`ENABLED_REPORTS=bcl2barcode,call_ready_tar,call_ready_rna,call_ready_wgs,sample_swaps,single_lane_tar,single_lane_rna,single_lane_wgs,single_lane_cfmedip`

UHN can instead set ENABLED_REPORTS to just this report they need, so the template stays scoped to their use case.

avarsava (approved):

slazicoicr (approved):

PR #451, opened 2026-08-18
Configurable list of report directories

avarsava:
The oicr-reports retain their many, many calls to df_manipulation, and df_manipulation remains in place because sidebar_utils requires it. It's less clean than it could be if we wanted the main repo to be truly just the dashi core, but it comes up. Third party reports could use df_manipulation if they wanted, I guess, or their own modules.

- [ ] Updates CHANGELOG.md
- [pending approval of the code changes] Updates dev docs if applicable


application/dash_application/pages.py
```diff
@@ -71,22 +36,49 @@ def f(_):
 
     return f
 
+# Get the installed pages
+all_reports = []
+pages = []
+_report_sources = os.getenv("REPORT_SOURCES")
+_enabled_env = os.getenv("ENABLED_REPORTS").split(",")
```
slazicoicr, line 43: This is optional, so can be `None` and will crash.

application/dash_application/pages.py
```diff
@@ -71,22 +36,49 @@ def f(_):
...
-        ))
+for source in _report_sources.split(","):
+    sys.path.append(source)
+    for (dirpath, _, filenames) in os.walk(source):
+        for filename in filenames:
+            if filename.endswith(".py"):
+                try:
+                    module = importlib.import_module(os.path.join(dirpath, filename))
+                    name = getattr(module, "page_name").replace("-", "_")
+                    if name in _enabled_env:
```
slazicoicr, line 56: `_enabled_env` uses `split(,)`, which will leave any white spaces that people usually put between commas, causing `name` not to match.

application/dash_application/pages.py
```diff
@@ -71,22 +36,49 @@ def f(_):
...
+                        lambda: "Unknown version",
+                        filename,
+                        lambda x: None,
+                    ))
 
+_enabled_env = os.getenv("ENABLED_REPORTS")
+if _enabled_env:
+    _requested = [n.strip() for n in _enabled_env.split(",") if n.strip()]
+    for _name in _requested:
+        if _name not in all_reports:
```
slazicoicr, line 79: Isn't this duplicated logic from `if name in _enabled_env: all_reports.append(name)`. `all_reports` will contain only reports in `_enabled_env` and this then checks if reports in `_enabled_env` are in `all_reports`.

avarsava:
Does it make sense, then, to remove ENABLED_REPORTS if the administrator is now in control of what report modules dashi sees?

